PR #1, opened 2017-04-08
Enum Editor Control

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/EnumEditorControl.cs
```diff
@@ -0,0 +1,83 @@
...
+using Foundation;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class EnumEditorControl : PropertyEditorControl
+	{
+		NSComboBox ComboBoxEditor;
+
+		dynamic EnumEditorViewModel => ViewModel;
```
ermau: Let's not bring the DLR into this. We can make `EnumEditorControl` generic and initialize it with `MakeGenericType` and `Activator`.

CartBlanche: @ermau Why? Won't there be equal overheads whether the DLR finds the right object or whether we test for it and create the appropriate object. This way the code is much *much*  simpler to read.

CartBlanche: Also my generic fu may not be too hot but AFACT you can cast the generic CreateInstance as 
`(PropertyViewModel) Activator.CreateInstance (type, property, this.editors);`
but you couldn't do 
`var epvm = (EnumPropertyViewModel<>)Activator.CreateInstance (type, property, this.editors);`
or 
`var epvm = (EnumPropertyViewModel<type>)Activator.CreateInstance (type, property, this.editors);`
Which I think would mean that to access the ValueName property we'd have to jump through a  few hoops by calling...
`var epvm = (PropertyViewModel)Activator.CreateInstance (type, property, this.editors);
var epvmType = epvm.GetType();
var valueNameProperty = epvmType.GetProperty("ValueName");
var valueNameValue = (string)valueNameProperty.GetValue (epvm);`

Am I missing a simpler way of doing this? That seems quite convoluted when we can leverage the power of the DLR and just call 
`EnumEditorViewModel.ValueName`
rather than the forementioned hoops.
Running the code here doesn't seem that slow. Its not like we are on i386 processors these days :)

Xamarin.PropertyEditing.Mac/Controls/EnumEditorControl.cs
```diff
@@ -9,11 +9,12 @@
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	internal class EnumEditorControl : PropertyEditorControl
+	internal class EnumEditorControl<T> : PropertyEditorControl
```
CartBlanche: @ermau I think this change to make the control generic, as per @alanmcgovern 's suggestion is a more elegant and cleaner fix than jumping through the Activator etc hoops.

PR #2, opened 2017-04-11
Add Property Search Capabilites...

CartBlanche:


Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -164,6 +204,38 @@ private PropertyViewModel GetViewModel (IPropertyInfo property)
...
+				}
+				return cachedProperties;
+			}
+		}
+
+		IEnumerable<IGrouping<string, PropertyViewModel>> cachedPropertiesGroupBy;
+		public IEnumerable<IGrouping<string, PropertyViewModel>> CachedPropertiesGroupBy {
+			get {
+				if (cachedPropertiesGroupBy == null) {
+					cachedPropertiesGroupBy = CachedProperties.GroupBy (arg => arg.Category);
```
ermau: This is just a cache of the lazy grouping, it's going to regroup each enumeration of it anyway, we need to capture the results into a list.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -164,6 +204,38 @@ private PropertyViewModel GetViewModel (IPropertyInfo property)
...
+			get {
+				if (cachedProperties == null) {
+					cachedProperties = Properties.ToList ();
+				}
+				return cachedProperties;
+			}
+		}
+
+		IEnumerable<IGrouping<string, PropertyViewModel>> cachedPropertiesGroupBy;
+		public IEnumerable<IGrouping<string, PropertyViewModel>> CachedPropertiesGroupBy {
```
ermau: The consuming UI code doesn't need to know this is cached, no need to leak implementation details into the API naming.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -164,6 +204,38 @@ private PropertyViewModel GetViewModel (IPropertyInfo property)
...
 
+		void FilterData ()
+		{
+			this.properties.Clear ();
+
+			UpdateProperties ();
+		}
+
+		List<PropertyViewModel> cachedProperties;
+		public List<PropertyViewModel> CachedProperties {
```
ermau: This is redundant, we don't need a copy of the list with no differences.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -164,6 +204,38 @@ private PropertyViewModel GetViewModel (IPropertyInfo property)
 			return new StringPropertyViewModel (property, this.editors);
 		}
 
+		void FilterData ()
+		{
+			this.properties.Clear ();
+
+			UpdateProperties ();
```
ermau: The `Properties` list should expose a second, filtered list of the properties, but we can build a filtered list without rebuilding the original property list as `UpdateProperties` does.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -164,6 +204,38 @@ private PropertyViewModel GetViewModel (IPropertyInfo property)
...
+				if (cachedPropertiesGroupBy == null) {
+					cachedPropertiesGroupBy = CachedProperties.GroupBy (arg => arg.Category);
+				}
+				return cachedPropertiesGroupBy;
+			}
+		}
+
+		void PropertiesViewModel_PropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
+		{
+			cachedProperties = null;
```
ermau: We don't need to invalidate the cache any time _any_ property changes. What we want is the original property list built by `UpdateProperties` stored safely, and then a second cached list that is a filtered view of the original list. `Properties` should expose The filtered list (and grouped list) are what we want to cache and invalidate whenever the original list changes, or the filter changes.

Xamarin.PropertyEditing.Tests/PanelViewModelTests.cs
```diff
@@ -415,7 +413,45 @@ public void PropertiesListSelectedItemResetStopsListening ()
...
+			var vm = new PanelViewModel (provider);
+			vm.SelectedObjects.Add (obj);
+
+			Assert.That (vm.Properties, Is.Not.Empty);
+			Assume.That (vm.Properties.Count, Is.EqualTo (2));
+			vm.FilterText = "sub";
+			Assume.That (vm.Properties.Count, Is.EqualTo (1));
+			vm.FilterText = string.Empty;
+			Assume.That (vm.Properties.Count, Is.EqualTo (2));
 		}
```
ermau: Arrange tests?

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
CartBlanche: I've tried various ways of caching and they just aren't working. It may be how NS*View populates. The following works, but obviously doesn't in cached. 

`public IReadOnlyList<PropertyViewModel> Properties { 
			get {
				if (!IsFiltering) {
					return this.properties;
				}
				else {
					return this.properties.Where ((arg) => arg.Property.Name.StartsWith (filterText, StringComparison.InvariantCultureIgnoreCase)).ToList ();
				}
			}
		}`

CartBlanche:
As we supposedly have a team waiting to use this, my suggestion would be to accept the PR for a working solution, that we know is not yet optimised and we raise an issue to do an optimisation pass after more controls are completed and before an official 1.0 release it done.

ermau:
The filter and arrange are attached to the bottom, if I resize the window vertically, they move behind other controls.

CartBlanche:
@ermau Nicely spotted! Latest commit should fix that resize issue.

Xamarin.PropertyEditing.Tests/PanelViewModelTests.cs
CartBlanche: @ermau Did you mean new tests or that re-arrange the tests already there?

PR #3, opened 2017-04-12
Mac Point Property ViewModel and Editor

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/PointEditorControl.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Drawing;
+using AppKit;
+using CoreGraphics;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class PointEditorControl : PropertyEditorControl
+	{
+		internal NSTextField XEditor { get; set; }
```
ermau: These don't actually need to be internal do they?

ermau:
Needs a rebase, then 👍

Xamarin.PropertyEditing.Mac/Controls/PointEditorControl.cs
CartBlanche: @ermau they don't have to be, but I was trying to be consistent with how the previous controls set up their NS* controls. If we change it here, then I think we should do a pass and remove them in all the controls.

ermau: Generally I don't like doing commits to fix formatting or stuff like this because it makes the history harder to get useful information from. I prefer to squash it into the original commit before it goes in, or to fix it when there's another change touching that line, at least ideally.

PR #5, opened 2017-04-19
Rectangle editor

CartBlanche:
Adding the possibility to support multiple 4 point objects
1st version is purely CGRect on Mac.
screen shot 2017-04-18 at 22 44 36


Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
```diff
@@ -0,0 +1,57 @@
...
+
+			WidthEditor.Frame = new CGRect (205, 0, 50, 20);
+
+			HeightLabel.Frame = new CGRect (260, -5, 45, 20);
+			HeightLabel.Value = "Height:";
+
+			HeightEditor.Frame = new CGRect (305, 0, 50, 20);
+		}
+
+		protected override void UpdateModelValue ()
```
ermau: I'd rather see an override in the final implementation that handles this than looking up and dynamically invoking multiple properties for every instance of this property.

Xamarin.PropertyEditing.Mac/Controls/BaseRectangleEditorControl.cs
```diff
@@ -0,0 +1,117 @@
...
+using AppKit;
+using CoreGraphics;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal abstract class BaseRectangleEditorControl<T> : PropertyEditorControl
+		where T : struct
+	{
+		internal NSTextView XLabel { get; set; }
```
ermau: I thought we weren't going to make these internal anymore? They should be protected and private set, nothing outside the controls should need to access them.

CartBlanche: I'll make them protected so they can be customised in the inherited classes.

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
CartBlanche: So you mean and Override with a specific types per class, to avoid the lookup? eg. `RectangleEditorControl<CGRect>` or `RectangleEditorControl<RectangleF>`so that in the UpdateModalValue we use ViewModel.Value which is already of CGRect or RectangleF respectively?

ermau: Right, just like we have ViewModels for each, they should be relatively trivial in size.

ermau: Alternatively and probably better, there aren't enough rectangle types to worry about, a type-based cache of delegates to the getters is probably enough.

CartBlanche: I've only just seen your last comment @ermau and already removed the generics. Are you referring to the VM's getter?

ermau: No, the getter of the individual properties on the Rectangle representation. I wasn't saying remove the generics, I was saying do something like `CGRectEditor : BaseRectEditor<CGRect>`, exactly like you said: 

> eg. RectangleEditorControl<CGRect> or RectangleEditorControl<RectangleF>so that in the UpdateModalValue we use ViewModel.Value which is already of CGRect or RectangleF

PR #6, opened 2017-04-21
Initial Color Editor Control using NSColorWell.

CartBlanche:
screen shot 2017-05-04 at 16 29 13

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/ColorEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/ColorEditorControl.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Collections;
+using System.Diagnostics;
+using AppKit;
+using CoreGraphics;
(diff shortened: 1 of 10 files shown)


alanmcgovern:
This control is already implemented in our codebase so we should re-use it. We have a much better popover based impl there which we can transplant into this repo!

CartBlanche:
@alanmcgovern was it using NSColorWell or something else?

CartBlanche:
from @viniciusjarina - https://github.com/xamarin/md-addins/blob/master/Xamarin.Designer.iOS/MonoTouch.Design.Client.Mac/NativePropertyEditing/ColorPropertyEntry.cs#L505

ermau:
This is such a complex piece, I'm going to close it for now until we see what we're getting from design as hopefully we can do all our changes at once. Furthermore, we want our controls to always deal with the shared value type, which in this case would be the System.Drawing colors types, not the native Cocoa ones.

PR #12, opened 2017-05-17
Use new ported NumericSpinEditor control

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+using AppKit;
+using CoreGraphics;
+using Foundation;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public class NumericSpinEditor : NSView
+	{
+		protected NumericTextField numericEditor;
+		protected NSStepper stepper;
```
alanmcgovern: These should not be exposed outside of this class. The values stored in them must be kept in sync otherwise Bad Things happen. If we publicly expose them we can't guarantee they'll always be in sync

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+		}
+
+		public string PlaceholderString {
+			get { return ((NSTextFieldCell)numericEditor.Cell).PlaceholderString; }
+			set { ((NSTextFieldCell)numericEditor.Cell).PlaceholderString = value; }
+		}
+
+		public NSStepper Stepper {
+			get { return stepper; }
+		}
```
alanmcgovern: I don't think we should expose this publicly for the same reason as above ;)

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+			get { return numericEditor.NumericMode; }
+			set {
+				numericEditor.NumericMode = value;
+				Reset ();
+			}
+		}
+
+		public string StringValue {
+			get { return numericEditor.StringValue; }
+			set { numericEditor.StringValue = value; }
```
alanmcgovern: We should not have a public setter that *only* updates the string value. Setting the string value here does not update the value stored in `stepper`!

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+			get { return numericEditor.BaselineOffsetFromBottom; }
+		}
+
+		public int Digits {
+			get { return (int)formatter.MaximumFractionDigits; }
+			set { formatter.MaximumFractionDigits = value; }
+		}
+
+		public virtual double Value {
+			get { return stepper.DoubleValue; }
```
alanmcgovern: Similarly we need to be careful about exposing the raw double value.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+			stepper.DoubleValue = CoerceValue (FieldValidation.FixInitialValue (value, Value.ToEditorString ()).ToEditorDouble ());
+			numericEditor.StringValue = FieldValidation.RoundDoubleValue (stepper.DoubleValue.ToEditorString (), NumericMode == ValidationType.Decimal ? FieldValidation.DefaultXcodeMaxRoundDigits : 0);
+		}
+
+		public virtual void SetValue (double value)
+		{
+			//Regulates maximun and minium out of range
+			stepper.DoubleValue = CoerceValue (value);
+			numericEditor.StringValue = FieldValidation.RoundDoubleValue (stepper.DoubleValue.ToEditorString (), NumericMode == ValidationType.Decimal ? FieldValidation.DefaultXcodeMaxRoundDigits : 0);
+		}
```
alanmcgovern: I don't like how we have two `SetValue` overrides, both of which are pretty complicated. Any control which overrides one must also override the second, and we have had bugs in the past where that was not done. Can we simplify this? Also, does `CoerceValue` also need to be public and overridable?

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+				NSLayoutConstraint.Create (stepper, NSLayoutAttribute.Right, NSLayoutRelation.Equal, control, NSLayoutAttribute.Left, 1, -6)
+			});
+			// tODO RemoveConstraint (rightSideConstraint);
+		}
+
+		public void SimulateTextInput (string input)
+		{
+			numericEditor.StringValue = input;
+			OnFieldEditingEnded (numericEditor, EventArgs.Empty);
+		}
```
alanmcgovern: This was added as a way to simplify unit testing. Do we actually still need it?

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -0,0 +1,287 @@
...
+
+		public void GoUp ()
+		{
+			stepper.DoubleValue = stepper.DoubleValue + IncrementValue;
+			OnStepperActivated (stepper, EventArgs.Empty);
```
alanmcgovern: We should ensure the `Stepper` value and `NumericTextField` value are in sync before any events visible outside of this class are invoked, otherwise people could see inconsistent values!

Xamarin.PropertyEditing/StringConversionExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+		/// Handle versions like '3035' and also '3035.24'. By default System.Version blows up if there's no '.'
+		/// </summary>
+		public static Version ToVersion (this string value)
+		{
+			if (string.IsNullOrEmpty (value))
+				return new Version ();
+			if (value.Contains ("."))
+				return Version.Parse (value);
+			return new Version (int.Parse (value), 0);
+		}
```
alanmcgovern: Do we need all of these extension methods ported over?

alanmcgovern:
A lot of the issues I picked out are pre-existing issues in the control. We can definitely do a bit of cleanup to remove all the sources of issues and bugs that affect us today!

It might be worth having a chat with Eric as well to see what he thinks of this... but it may make sense to expose a public type like:
```
struct NumericValue
{
	// Normal values like 1, 200, 400 would be stored as:
	// Numerator = 100, Denominator = 1
	//
	// Ratio values such as 1:3 or 5/6 would be stored as
	// Numerator = 1, Denominator = 3 etc.
	public double Numerator { get; }
	public double Denominator { get; }

	// This can be either '/' or ':'
	public char RatioSeparator { get; }

	public string ValueString {
		get {
			if (Denominator == 1)
				return Numerator.ToString ();
			else
				return Numerator + "/" + Denominator;
		}
	}
}
```
Then we could more easily keep track of ratio values versus regular values.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
CartBlanche: @alanmcgovern the broken syncing is down to the fact that SetValue, isn't being used whereever the value is set/changed. Using that method appears to fix the syncing issue.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
CartBlanche: Agreed this should just be 1 method.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
CartBlanche: Fixed locally.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
CartBlanche: @alanmcgovern why worrying? Do you think the only access should be via StringValue?

alanmcgovern: I think we should not have multiple public ways of setting values when each way does something different. This control is literally impossible for someone to use with the way it's currently written. These kinds of bugs have been the bane of my life for the last few months :( See my main comment at the end. There's definitely got to be a better way of handling this other than trying to publicly expose the raw string value and double value :)

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
```diff
@@ -30,24 +30,27 @@ public BasePointEditorControl ()
...
-			YEditor.Activated += OnInputUpdated;
-			YEditor.EditingEnded += OnInputUpdated;
+			YEditor.ValueChanged += OnInputUpdated;
 
 			AddSubview (XLabel);
 			AddSubview (XEditor);
 			AddSubview (YLabel);
 			AddSubview (YEditor);
+
+            this.DoConstraints ( new[] {
```
alanmcgovern: tabs/spaces here

Xamarin.PropertyEditing.Mac/Controls/CGRectEditorControl.cs
```diff
@@ -29,12 +29,19 @@ public CGRectEditorControl ()
 
 			HeightEditor.Frame = new CGRect (150, 0, 50, 20);
 
+            this.DoConstraints ( new[] {
```
alanmcgovern: tabs/spaces

Xamarin.PropertyEditing/StringConversionExtensions.cs
alanmcgovern: How many of these methods can we remove? Let's remove everything we don't use right now!

CartBlanche:
@alanmcgovern do you think it is worth investing the time to change the underlying workings to support the NumericValue struct?

alanmcgovern:
I'd say we could do that as a phase 2 in a future PR. Not worth doing it now i think :)

alanmcgovern:
Good memory though, i'd forgotten i suggested that!

CartBlanche:
Agreed on phase 2.

CartBlanche:
@alanmcgovern Based on having a common numerator/denominator core for iOS, I think I'll make the change to support the NumericValue struct. Thoughts?

PR #13, opened 2017-05-18
Initial Accessibility support for our Mac editors.

CartBlanche:


ermau:
I haven't yet checked the macOS narrator, but I know Windows' tells you what kind of control you're dealing with. Does macOS' not? Seems like we're adding a lot of text here that the narrator should be doing for us.

PR #16, opened 2017-05-18
Windows accessibility improvements

ermau:
- Adds narrator support for a number of controls
 - Improves keyboard navigation

### DIFF
--- a/Xamarin.PropertyEditing.Windows/NumericUpDownControl.cs
+++ b/Xamarin.PropertyEditing.Windows/NumericUpDownControl.cs
@@ -62,6 +62,11 @@ namespace Xamarin.PropertyEditing.Windows
 		: Control
 		where T : struct, IComparable<T>
 	{
+		static NumericUpDownControl ()
+		{
+			FocusableProperty.OverrideMetadata (typeof (NumericUpDownControl<T>), new FrameworkPropertyMetadata (false));
+		}
+
 		public NumericUpDownControl ()
 		{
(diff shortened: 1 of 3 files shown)


Xamarin.PropertyEditing.Windows/NumericUpDownControl.cs
```diff
@@ -62,6 +62,11 @@ internal abstract class NumericUpDownControl<T>
 		: Control
 		where T : struct, IComparable<T>
 	{
+		static NumericUpDownControl ()
+		{
+			FocusableProperty.OverrideMetadata (typeof (NumericUpDownControl<T>), new FrameworkPropertyMetadata (false));
```
CartBlanche: Don't we want the Focusable to be *true* in this case, while PropertyEditorControl is *false*?

alanmcgovern: I could have forgotten my focusable rules so forgive me if i get it wrong ;) The numeric control itself should not be focusable or a tab stop because it does nothing in and of itself. The `up` button, `down` button and the entry textbox contained inside it should be focusable/tabstops though

CartBlanche: @alanmcgovern Ah I didn't look close enough, I thought the NumericUpDown was both the input box and up/down controls, hence why I thought it needed focus.

alanmcgovern: aah, gotcha!

PR #22, opened 2017-06-14
Changes to support correct tabbing through TableView cells

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/UnfocusableTextView.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using AppKit;
+using CoreGraphics;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	class UnfocusableTextView : NSTextView
```
ermau: Maybe something a bit shorter and more consistent? `NSLabel` ?

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -124,8 +124,10 @@ private void UpdateProperties (IObjectEditor[] removedEditors = null, IObjectEdi
 			}
 
 			foreach (IPropertyInfo property in newSet) {
-				this.properties.Add (GetViewModel (property));
```
ermau: Unnecessary and inconsistent change.

Xamarin.PropertyEditing.Mac/Controls/UnfocusableTextView.cs
CartBlanche: NS-prefix is NeXtStep. I don't think we should imply out control is Native with the NS-prefix. With code completion so prevalent I don't think having long descriptive class names is a major issue for developers.

PR #23, opened 2017-06-19
[build] Simplify our VSTS/wrench integration 

ermau:


alanmcgovern:
I can't merge on a phone :p

We can keep tweaking this and updating it as we go, but this is good enough to merge!

PR #24, opened 2017-06-19
Mac search and arrange

ermau:
This implements filter searching and arranging on Mac. It does temporarily break the tabbing fix introduced just before as that was based on a regular `NSTableView` and not an `NSOutlineView`. We'll fix that separately in the interest of getting this in.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using CoreGraphics;
 using Foundation;
 using AppKit;
 using Xamarin.PropertyEditing;
@@ -16,6 +17,9 @@ namespace Xamarin.PropertyEditing.Mac
 {
 	public partial class PropertyEditorPanel : AppKit.NSView
 	{
+		NSSearchField propertyFilter;
+		NSComboBox propertyFilterMode;
+
 		public PropertyEditorPanel ()
 		{
 			Initialize ();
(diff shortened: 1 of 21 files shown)


Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
```diff
@@ -129,11 +115,12 @@ void Initialize ()
...
 				AutoresizingMask = NSViewResizingMask.WidthSizable,
 				SelectionHighlightStyle = NSTableViewSelectionHighlightStyle.None,
+				RowHeight = 24,
 			};
 
-			// create columns for the panel
-			NSTableColumn propertiesList = new NSTableColumn ("PropertiesList") { Title = PropertyListTitle };
-			NSTableColumn propertyEditors = new NSTableColumn ("PropertyEditors") { Title = PropertyEditorTitle };
+			// TODO: localize
+			NSTableColumn propertiesList = new NSTableColumn (PropertyListColId) { Title = "Properties" };
```
CartBlanche: I really think this should be singular. The column name refers to a single Property.

Xamarin.PropertyEditing.Mac/PropertyTableDataSource.cs
```diff
@@ -3,34 +3,80 @@
...
-using Xamarin.PropertyEditing.Mac.ViewModels;
 using Xamarin.PropertyEditing.ViewModels;
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	public class PropertyTableDataSource : NSOutlineViewDataSource
+	internal class PropertyTableDataSource
+		: NSOutlineViewDataSource
 	{
-		internal MacPanelViewModel ViewModel { get; private set; }
```
CartBlanche: I don't understand why the MacPanelViewModel class has been removed. It encapsulated all the logic for GetChildrenCount, ItemExpandable, GetChildObject and made the code here tidier.

ermau: There wasn't any real reason for it. There's no reuse case, it simply moves code from where it was actually being used to another class and it's not really complicated enough to warrant tests.

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
```diff
@@ -15,5 +11,177 @@ public PanelViewModel (IEditorProvider provider)
...
+			}
+		}
+
+		private bool MatchesFilter (PropertyViewModel vm)
+		{
+			if (String.IsNullOrWhiteSpace (FilterText))
+				return true;
+
+			string filter = FilterText.ToLower ();
+			if (ArrangeMode == PropertyArrangeMode.Category && vm.Category != null && vm.Category.ToLower().Contains (filter)) {
```
alanmcgovern: The `ToLower` here has me worried! What's the expected behaviour in French, English and Turkish locales. Turkey has 4 different variants of 'i' so this is unlikely to have the desired effect there.

http://www.i18nguy.com/unicode/turkish-i18n.html
https://blog.codinghorror.com/whats-wrong-with-turkey/

And/or google a few more links :p

alanmcgovern: We should review all `ToLower`, `ToUpper` calls, and also be careful about anything that deals with culture specific number formatting.

alanmcgovern: (also `StringComparison.CurrentCultureIgnoreCase` might be handy for this kind of comparison )

Xamarin.PropertyEditing/ObservableLookup.cs
```diff
@@ -0,0 +1,336 @@
...
+			get
+			{
+				if (key == null)
+					return this.nullGrouping;
+
+				ObservableGrouping<TKey, TElement> grouping;
+				if (this.groupings.TryGetValue (key, out grouping))
+					return grouping;
+
+				return new TElement[0];
```
alanmcgovern: `Enumerable.Empty<TElement>` or `Array.Empty<TElement>` should be preferred generally

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
ermau: Ah yeah, this was `StartsWith` with oridinal compare, missed it when I changed it over.

alanmcgovern: Does `Ordinal` do the right thing for the 4 variants of 'i' that the turkish locale has? I think we'll actually need the culture aware insensitive comparison in order to handle that!

PR #28, opened 2017-06-27
Added RowHeight support and refactor CGRect to be multi-line.

CartBlanche:


Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -125,6 +125,14 @@ public override void ItemDidCollapse (NSNotification notification)
 				this.dataSource.DataContext.SetIsExpanded (group.Key, isExpanded: false);
 		}
 
+		public override nfloat GetRowHeight (NSOutlineView outlineView, NSObject item)
+		{
+			var facade = (NSObjectFacade)item;
+			var vm = facade.Target as PropertyViewModel;
+			var group = facade.Target as IGroupingList<string, PropertyViewModel>;
+			return  (group == null) ? vm.RowHeight : group.ElementAtOrDefault(1).RowHeight;
```
ermau: If the group is empty, this will NRE. If there's cases of "this can't be null at this stage", don't use safeguard methods like `OrDefault`. Additionally, `IGroupingList` has an indexer, always prefer that over `ElementAt`:
```csharp
internal interface IGroupingList<TKey, TElement>
		: IGrouping<TKey, TElement>, IReadOnlyList<TElement>
```

Xamarin.PropertyEditing.Mac/Controls/EnumEditorControl.cs
```diff
@@ -76,16 +74,57 @@ protected override void UpdateErrorsDisplayed (IEnumerable errors)
...
+			if (!dataPopulated) {
+				if (EnumEditorViewModel.IsFlags) {
+					var top = 0;
+					foreach (var item in EnumEditorViewModel.PossibleValues) {
+						var BooleanEditor = new NSButton (new CGRect(0, top, 200, 24)) { TranslatesAutoresizingMaskIntoConstraints = false };
+						BooleanEditor.SetButtonType (NSButtonType.Switch);
+						BooleanEditor.Title = item.Key;
+						BooleanEditor.State = item.Value ? NSCellStateValue.On : NSCellStateValue.Off;
+						BooleanEditor.Activated += BooleanEditor_Activated;
+						BooleanEditor.Enabled = false; // TODO Remove this line once EnumEditorViewModel.Value is updated correctly.
```
ermau: I'd rather it was complete before it was merged instead of piecemealing it in.

Xamarin.PropertyEditing/ViewModels/EnumPropertyViewModel.cs
```diff
@@ -11,16 +13,25 @@ internal class EnumPropertyViewModel<TValue>
...
+			}
 		}
 
 		public bool IsFlags
 		{
 			get;
 		}
 
-		public IReadOnlyList<string> PossibleValues
+		public IReadOnlyDictionary<string, bool> PossibleValues
```
ermau: This change would break Windows. Whether or not a flag is set doesn't have anything to do with what values are available to set, flags set/unset should be separate.

PR #29, opened 2017-06-27
Added RowHeight support and refactored CGRect to be multi-line.

CartBlanche:
By creating the editors in the RowHeight we can get the editor's height straight away. If the editor is already created, event better.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/CGRectEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/CGRectEditorControl.cs
@@ -9,25 +9,27 @@ namespace Xamarin.PropertyEditing.Mac
 		public CGRectEditorControl ()
 		{
 			// TODO localize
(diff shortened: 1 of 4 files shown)


Xamarin.PropertyEditing.Mac/Controls/CGRectEditorControl.cs
```diff
@@ -48,6 +48,8 @@ protected override void UpdateModelValue ()
 			YEditor.StringValue = ViewModel.Value.Y.ToString ();
 			WidthEditor.StringValue = ViewModel.Value.Width.ToString ();
 			HeightEditor.StringValue = ViewModel.Value.Height.ToString ();
+
+			ViewModel.RowHeight = 48;
```
CartBlanche: @ermau The Heights are being set by the View. The VM is being passed to the GetRowHeight method via the NSObjectFacade. Unless we also try to pass the view as another property of the NSObjectFacade.

PR #30, opened 2017-06-27
Start theming

ermau:


Xamarin.PropertyEditing.Windows/Resources.xaml
```diff
@@ -131,22 +131,23 @@
...
+							<ColumnDefinition SharedSizeGroup="Label" />
+							<ColumnDefinition />
+						</Grid.ColumnDefinitions>
+						<Label Name="Label" Target="{Binding ElementName=CheckBox,Mode=OneTime}" Content="{TemplateBinding Label}" />
+						<CheckBox Name="CheckBox" AutomationProperties.LabeledBy="{Binding ElementName=Label,Mode=OneTime}" Grid.Column="1" IsChecked="{Binding Value}" IsEnabled="{Binding Property.CanWrite,Mode=OneTime}" VerticalAlignment="Center" />
+					</Grid>
+				</ControlTemplate>
+			</Setter.Value>
+		</Setter>
+	</Style>
```
alanmcgovern: In future can we try to correct whitespace changes by squashing commits before merging. Secondly if we do have to fix it up can we avoid mixing whitespace changes and functional changes?

The diff here is the removal of the `Content` binding, but it's not easy to spot unless you use `?w=1` in the url :p

PR #37, opened 2017-07-11
Make Get/Set async

ermau:


Xamarin.PropertyEditing/Reflection/ReflectionEnumPropertyInfo.cs
```diff
@@ -65,7 +65,7 @@ public override void SetValue<TValue> (object target, TValue value)
 			}
 		}
 
-		public override TValue GetValue<TValue> (object target)
+		public override async Task<TValue> GetValueAsync<TValue> (object target)
```
sandyarmstrong: These methods are `async` but lack any `awaits`. Is that intentional? They'll run synchronously.

Xamarin.PropertyEditing/Reflection/ReflectionPropertyInfo.cs
```diff
@@ -61,7 +62,7 @@ public virtual void SetValue<T> (object target, T value)
 			this.propertyInfo.SetValue (target, realValue);
 		}
 
-		public virtual T GetValue<T> (object target)
+		public virtual async Task<T> GetValueAsync<T> (object target)
```
sandyarmstrong: Same question here on not really being `async`. Keep in mind I don't know this code at all. :-)

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -153,27 +157,33 @@ private bool SetCurrentValue (ValueInfo<TValue> newValue)
...
-				AggregateException aggregate = ex as AggregateException;
-				if (aggregate != null) {
-					aggregate = aggregate.Flatten ();
-					ex = aggregate.InnerExceptions[0];
+			using (await AsyncWork.RequestAsyncWork (this)) {
+				try {
+					Task[] setValues = new Task[Editors.Count];
+					int i = 0;
+					foreach (IObjectEditor editor in Editors) {
+						setValues[i++] = editor.SetValueAsync (Property, newValue);
```
sandyarmstrong: Shouldn't this be `await`ed?

alanmcgovern: look 2 lines further down :p

Xamarin.PropertyEditing/Reflection/ReflectionEnumPropertyInfo.cs
ermau: That's intentional. The `Reflection/` code is all just for testing, so it doesn't need to be all that great, this was just the quickest way of going from not building to building instead of sprinkling `Task.FromResult` everywhere.

ermau: And to clarify, it's completely ok for these to run synchronously, regardless of whether it's supposed to be a good implementation or not. We're more enabling providers (like yours) to be async if they need to be, not necessarily saying they _should_ be.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
sandyarmstrong: Whoops!

Xamarin.PropertyEditing/IObjectEditor.cs
```diff
@@ -51,9 +52,27 @@ public interface IObjectEditor
...
+		 * Anything that needs to deal with user input will need to block its UI. We simply can't reason about the user
+		 * changing values while waiting for previous changes, because they value they're changing may change underneath
+		 * them or become invalid as a result of a prior pending change.
+		 */
+
+		Task SetValueAsync<T> (IPropertyInfo property, ValueInfo<T> value, PropertyVariation variation = null);
 
 		/// <exception cref="ArgumentNullException"><paramref name="property"/> is <c>null</c>.</exception>
-		ValueInfo<T> GetValue<T> (IPropertyInfo property, PropertyVariation variation = null);
+		Task<ValueInfo<T>> GetValueAsync<T> (IPropertyInfo property, PropertyVariation variation = null);
```
sandyarmstrong: Is it worth adding cancellation to these APIs? Maybe good to put them in the interface even if actually using them comes later?

ermau: Possibly. There are scenarios in which we could cancel, but whether it's beneficial really depends on your stuff.

ermau: Bringing it back from offline discussion: I don't really think we buy ourselves enough to add it and implicitly encourage people to implement cancellation.

PR #39, opened 2017-07-12
Generic point editor control

CartBlanche:
Refactor to support multiple 2 point objects.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
@@ -9,49 +9,45 @@ using Xamarin.PropertyEditing.ViewModels;
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	internal class PointEditorControl : PropertyEditorControl
+	internal abstract class BasePointEditorControl<T> : PropertyEditorControl
+		where T : struct
 	{
+		internal UnfocusableTextField XLabel { get; set; }
 		internal NSTextField XEditor { get; set; }
+		internal UnfocusableTextField YLabel { get; set; }
 		internal NSTextField YEditor { get; set; }
 
 		public override NSView FirstKeyView => XEditor;
 		public override NSView LastKeyView => YEditor;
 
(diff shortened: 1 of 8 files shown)


Xamarin.PropertyEditing.Mac/Controls/PointEditorControl.cs
```diff
@@ -9,77 +9,31 @@
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	internal class PointEditorControl : PropertyEditorControl
+	internal class PointEditorControl : BasePointEditorControl<CGPoint>
```
ermau: If we have multiple subclasses that serve as shells for particular implementing types (`CGPoint`), we should name them accordingly.

Xamarin.PropertyEditing.Mac/Controls/SizeEditorControl.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+	internal class SizeEditorControl : BasePointEditorControl<CGSize>
+	{
+		public SizeEditorControl ()
+		{
+			XLabel.Frame = new CGRect (0, 0, 40, 24);
+			XLabel.StringValue = "Width:"; // TODO Localise
+
+			XEditor.Frame = new CGRect (45, 0, 50, 20);
+			XEditor.Activated += (sender, e) => {
```
ermau: We should consider the hooks in the base class that calls a virtual method to update the value as we should _also_ be updating value on blur.

Xamarin.PropertyEditing.Mac/Controls/PointEditorControl.cs
CartBlanche: So following a CGPointPointEditorControl and CGSizeEditorControl in this case?

Xamarin.PropertyEditing.Mac/Controls/SizeEditorControl.cs
CartBlanche: Can you clarify what you mean by 
> on blur
?

ermau: Loses focus

PR #41, opened 2017-07-14
Added .NET Point, Size and Rectangle editors.

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
```diff
@@ -0,0 +1,49 @@
...
+			WidthLabel.StringValue = "Width:";
+
+			WidthEditor.Frame = new CGRect (45, 0, 50, 20);
+
+			HeightLabel.Frame = new CGRect (105, 0, 45, 24);
+			HeightLabel.StringValue = "Height:";
+
+			HeightEditor.Frame = new CGRect (150, 0, 50, 20);
+
+			RowHeight = 48;
```
ermau: Shouldn't this be in the base?

PR #46, opened 2017-08-22
Make PropertyVariation equatable and expose ObservableCollectionEx

ermau:


Xamarin.PropertyEditing/PropertyVariation.cs
```diff
@@ -24,5 +25,37 @@ public string Name
...
+			if (ReferenceEquals (null, other))return false;
+			if (ReferenceEquals (this, other)) return true;
+			return String.Equals (Category, other.Category) && String.Equals (Name, other.Name);
+		}
+
+		public override bool Equals (object obj)
+		{
+			if (ReferenceEquals (null, obj)) return false;
+			if (ReferenceEquals (this, obj)) return true;
+			if (obj.GetType () != GetType ()) return false;
```
alanmcgovern: This override does a type check but the other one doesn't. Subclasses will be treated differently based on which method they go through. 

My preference on this is to implement the method as a 1-liner: 'Equals (obj as PropertyVariation)'

ermau: Addressed

PR #53, opened 2017-09-11
[Core/Win] Name object, selection type

ermau:


Xamarin.PropertyEditing.Tests/Xamarin.PropertyEditing.Tests.csproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\packages\NUnit3TestAdapter.3.8.0\build\net35\NUnit3TestAdapter.props" Condition="Exists('..\packages\NUnit3TestAdapter.3.8.0\build\net35\NUnit3TestAdapter.props')" />
```
alanmcgovern: Unrelated note - we should look into swapping this repo to `PackageReference` instead of normal `packages.config`

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -117,8 +202,44 @@ private void RemoveProperties (IEnumerable<PropertyViewModel> properties)
 			OnRemoveProperties (properties);
 		}
 
+		private async void SetObjectName (string value)
+		{
+			if (this.nameable == null)
+				return;
+
+			try {
+				await this.nameable.SetNameAsync (value);
```
alanmcgovern: Is there an ordering guarantee hidden in here to ensure the correct final value is used if SetObjectName is invoked multiple times?

ermau: No, it's up to the editor to guarantee (which happens for free if it's not actually async). It's likely that this will move when we introduce the async UI mechanics.

alanmcgovern:
#YOLO!

PR #54, opened 2017-09-14
Brush viewmodel

bleroy:


bleroy:
This should have parity with WPF, and solid brush covers simple colors as well. There are differences for Xamarin.Forms, and additional types, that would be required.

PR #56, opened 2017-09-19
Add support for color spaces to the solid brush view model.

bleroy:


Xamarin.PropertyEditing/ISolidBrushPropertyInfo.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+
+namespace Xamarin.PropertyEditing
+{
+	// TODO: Add support for predefined color values
+	public interface ISolidBrushPropertyInfo : IPropertyInfo
```
ermau: Generally stuff like this has been done as "light-up" interfaces for the individual feature (color spaces, in this case), rather than as a whole property type. This prevents us from having to make a bunch of these if other brush types end up needing color spaces and prevents platforms that don't need to from having to implement it anyway.

bleroy: Good point, thanks.

Xamarin.PropertyEditing/Reflection/ReflectionSolidBrushPropertyInfo.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Reflection;
+
+namespace Xamarin.PropertyEditing.Reflection
+{
+	public class ReflectionSolidBrushPropertyInfo : ReflectionPropertyInfo, IColorSpaced
+	{
+		public ReflectionSolidBrushPropertyInfo (PropertyInfo propertyInfo, IEnumerable<string> colorSpaces = null)
+			: base (propertyInfo)
+		{
+			ColorSpaces = colorSpaces.ToArray();
```
ermau: Shouldn't this be `colorSpaces?.ToArray();` ?

PR #57, opened 2017-09-21
Windows choice editor

ermau:


Xamarin.PropertyEditing.Windows/ChoiceControl.cs
```diff
@@ -1,58 +1,108 @@
...
+	{
+		public event EventHandler SelectedItemChanged;
 
 		public override void OnApplyTemplate ()
 		{
 			base.OnApplyTemplate ();
 
-			this.list = (ListBox)GetTemplateChild ("list");
-			this.list.SelectionChanged += OnListSelectionChanged;
+			ItemContainerGenerator.StatusChanged += (sender, args) => {
```
alanmcgovern: I assume we don't need to worry about the template changing, thus causing this method to be invoked twice and two event handlers being attached?

alanmcgovern:
Assuming there's no need to worry about my question, this is 👍

PR #59, opened 2017-09-26
Mock control models for consumption by standalone test apps

bleroy:
The new mock controls have specific implementations for WPF and AppKit buttons and controls. Those could complement or replace the reflection-based models, and can work cross-platform thanks to their lack of dependency on platform-specific libraries.


Xamarin.PropertyEditing.Mac.Standalone/Xamarin.PropertyEditing.Mac.Standalone.csproj
```diff
@@ -9,7 +9,7 @@
     <RootNamespace>Xamarin.PropertyEditing.Mac.Standalone</RootNamespace>
     <AssemblyName>Xamarin.PropertyEditing.Mac.Standalone</AssemblyName>
     <MonoMacResourcePrefix>Resources</MonoMacResourcePrefix>
-    <UseXamMacFullFramework>true</UseXamMacFullFramework>
```
ermau: Are we sure we don't need this?

Xamarin.PropertyEditing.Tests/MockPropertyInfo/MockPropertyInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+	{
+		public MockPropertyInfo (string name, string category = "", bool canWrite = true)
+		{
+			Name = name;
+			Category = category;
+			CanWrite = canWrite;
+		}
+
+		public string Name { get; }
+		public Type Type => typeof (T);
```
ermau: This doesn't work for enums, their representation type is the base type (int, short, etc) for the enum.

Xamarin.PropertyEditing.Windows.Standalone/MainWindow.xaml
```diff
@@ -18,8 +18,8 @@
 			<RowDefinition />
 		</Grid.RowDefinitions>
 
-		<Button Grid.Row="0" Grid.Column="0" Click="Button_Click">Select me</Button>
-		<Button Grid.Row="1" Grid.Column="0" Click="Button_Click">Select me 2</Button>
+		<local:MockedWpfButton Grid.Row="0" Grid.Column="0" Click="Button_Click">Select me (Mock)</local:MockedWpfButton>
```
ermau: Let's move to 3 buttons, two of the mocked type so we can still test same properties for multi-select.

bleroy: I'm thinking of adding a simpler mocked type that has only one property of each supported type. We could have two buttons of that. That would be simple mocked 1, simple mocked 2, full platform-specific mocked, reflection-based. What do you think?

ermau: That seems fine, that's kind of what I intended originally, but this was fine too.

bleroy: It's probably useful in the long run to keep at least one representation of a full control for accuracy and to get a good idea of the actual experience editing real controls, but as we work day to day on supporting more property types, having a more distraction-free experience will save us some time. I'll add it.

PR #60, opened 2017-10-04
Solid color brush editor

bleroy:


Xamarin.PropertyEditing/Drawing/CommonSolidBrush.cs
```diff
@@ -46,7 +47,8 @@ public override int GetHashCode ()
 			var hashCode = base.GetHashCode ();
 			unchecked {
 				hashCode = hashCode * -1521134295 + Color.GetHashCode ();
-				hashCode = hashCode * -1521134295 + ColorSpace.GetHashCode ();
+				if (ColorSpace != null)
+					hashCode = hashCode * -1521134295 + ColorSpace.GetHashCode ();
```
drewgillies: nit: please describe the significance of this value in a comment

ermau:
This is just the original control and does not match VS, there's a lot of work to still do.

bleroy:
@ermau Missing features and changes in order to get parity with the WPF solid brush editor:

* Add Hex ARGB textbox (S)
* Move the hue picker to a vertical band right to the shade picker (S)
* Add '> <' marks to the hue picker to show current selection (S)
* Add gradients to the textboxes showing previews of alterations to the color caused by changing each component (M)
* Add initial, current, and last color rectangles (S) - note: not sure what triggers a change of "initial"
* Add eyedropper button (unknown)
* Convert to resource button: if I'm not mistaken, this will need the support for resources to be added across the board

Let me know if I missed something

ermau:
There's also the fact that, in its standalone form (which is all we have right now), the actual inline editor is nothing more than a preview box. When clicked it makes a popup with the actual editor. See the OneNote design doc for more detailed info.

Convert to resource button? Do you mean the property button? Yes, that's not yet ready, we can skip that and the eye dropper for now.

bleroy:
One more from @ermau's comment:

* Make the whole editor a popup activated from clicking a read-only rectangle previewing the brush (M)

We'll skip the eyedropper and resource creation for now.

PR #63, opened 2017-10-06
Theme support for Property Editor Panel

CartBlanche:


Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Drawing;
+using AppKit;
+
+namespace Xamarin.PropertyEditing.Themes
+{
+	public class MacThemeManager : BaseThemeManager
+	{
+		public NSAppearance CurrentAppearance { get; private set; }
+
+		NSAppearance DarkAppearance = NSAppearance.GetAppearance (NSAppearance.NameVibrantDark);
```
ermau: There don't seem to be any of the colors defined by Vaclav present?

Xamarin.PropertyEditing.Windows/Themes/WinThemeManager.cs
```diff
@@ -0,0 +1,75 @@
...
+
+	public class WinThemeManager : BaseThemeManager
+	{
+		Dictionary<string, ResourceDictionary> themeList;
+
+		public WinThemeManager ()
+		{
+			themeList = new Dictionary<string, ResourceDictionary> ();
+
+			var asm = Assembly.GetExecutingAssembly ();
```
ermau: We can just hardcode a list of the themes and use `ResourceDictionary.Source` with a pack URI to import them instead of all this.

Xamarin.PropertyEditing/Themes/ThemeStyle.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Drawing;
+
+namespace Xamarin.PropertyEditing.Themes
+{
+	public class ThemeStyle
```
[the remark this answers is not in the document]
CartBlanche: This was previously used in the 1st version of the Mac implementation where I was setting up and setting the Control's Background etc.  It is no longer used, so I may remove it.

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
CartBlanche: As per my other comment regarding ThemeStyle, the 1st version I showed Vaclav used the ThemeStyle properties to style the control's backgrounds etc. But it wasn't 100% right and we would need to create custom controls, to achieve the styling. After speaking to him he said he wanted to styling to work like Xamarin Studio. This is exactly how XS achieves its styling, hence the change to dump the ThemeStyle code and use the NSAppearance mechanism above.

Xamarin.PropertyEditing.Windows/Themes/WinThemeManager.cs
CartBlanche: While I could find a way to change ResourceDictionary.Source in Application.Current.Resources.MergedDictionaries, however I could not find a way to change the ResourceDictionary.Source inside the ExecutingAssembly. If you have an easy way to do that, great as it would simplify things greatly. I'll ping you on Monday about it.

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
ermau: My point isn't about the mechanism used, but that the appearance selected does not match his colors selected in the design. I assume there's a way to define a custom appearance and that he isn't intending that you ignore his theme.

Xamarin.PropertyEditing.Windows/Themes/WinThemeManager.cs
ermau: You don't have to change the source, you remove the current theme's merged dictionary and add the other one.

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
CartBlanche: Custom themes are possible, but those were abandoned by the XS team due to the fact Apple have not provided a tool to create them and the unreliability of 3rd party tools, as told to me by Sevo Kukol from the XS team. 
Pinging Vaclav here... 
@vancura can you please confirm that we want the PropertyEditor to follow the Xamarin Studio manner of theming? This is what the Property Editor currently looks like following XS-Mac theming...

screen shot 2017-10-09 at 16 47 45
screen shot 2017-10-09 at 16 50 16

Xamarin.PropertyEditing.Windows/Themes/WinThemeManager.cs
```diff
@@ -0,0 +1,75 @@
...
+		protected override void SetTheme ()
+		{
+			if (themeList != null) {
+				// Making sure none of the themes are loaded, then we can just switch appropriately.
+				Application.Current.Resources.MergedDictionaries.Remove (themeList["vs.dark.baml"]);
+				Application.Current.Resources.MergedDictionaries.Remove (themeList["vs.light.baml"]);
+
+				switch (Theme) {
+					case PropertyEditorTheme.Dark:
+						Application.Current.Resources.MergedDictionaries.Add (themeList["vs.dark.baml"]);
```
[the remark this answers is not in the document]
CartBlanche: @ermau This ^^ is where I thought you meant the above property would be a bindable property. What did you mean here exactly?

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
vancura: Yes, please use XS theming. Colors defined in my comps are a part of future redesign, we'll get there, but for now use whatever we have in XS.

WRT your question about how we do it (/c @sevoku): we have a [file with all colors](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Styles.cs), so it would be best if Property Pad used colors from here. Maybe I am wrong, I don't exactly know what's the best approach code-wise, but I believe this is the best we can do (@sevoku might think something else).

If you use `Styles` from `MonoDevelop.Ide` we'll be able to change the colors later to fit the rest of redesign.

Makes sense?

vancura: I just have a comment about your two screenshots above; please take a look at the space between two columns: it seems you have both columns too close to each other, there needs to be a gap. Thanks!

CartBlanche: More like this...
screen shot 2017-10-10 at 16 18 10

ermau: @CartBlanche The spacing is defined in his document at 12pt.

> but I believe this is the best we can do (@sevoku might think something else).

We can't use that directly for a variety of reasons, we'll discuss code handling in Slack.

ermau: @vancura If we're going to use that color list, can you annotate the designs to call out what color titles each piece is using?

vancura: Yes, but not now – I am too busy with other stuff before coming to Boston on Saturday. I think I should be able to find some time for it next week, between meetings.

CartBlanche: @vancura this is a screen shot from the conversation we had about what it would look like using the NSAppearance Dark theme. which is what I implemented and PRed. 
screen shot 2017-10-11 at 15 33 11
Hence my confusion about why we now need to revert my original Styling change, which has lots of custom control implications, AFAICT. Maybe @sevoku can enlighten us further.

Maybe we should investigate custom NSAppearances again and get that working instead of the Styling property changes. Possibly fork an existing tool that does most of the custom NSAppearance heavy lifting and we can contribute fixes/enhancements so that the whole Apple community benefits from the enhanced tool, at least until Apple releases one??

I'll be pushing an update to fix the 12pixel separation issue shortly.

vancura: I based the comps on dark `NSAppearance` (and light too), or rather how bad it is – apparently Apple forgets about it, and all we have is just a buggy implementation with a ton of workarounds (@sevoku would tell stories).

I tried to mix colors and shapes to replicate the feeling of Apple's treatment, to make it look at usable. The target was to have as little work as possible when mixing with other controls by Apple. And also to have just minimum custom controls and use what they give us much as possible.

So yes, this is what I thought would be the end if everything works well. But of course, if we can't have it for any reason, I am open to discussion. It doesn't mean I like what Apple does, and it was just the fastest way and the most compatible.

PR #64, opened 2017-10-13
Initial Localisation pass.

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/BaseNumericEditorControl.cs
```diff
@@ -91,10 +91,10 @@ protected override void SetEnabled ()
 		protected override void UpdateAccessibilityValues ()
 		{
 			NumericEditor.AccessibilityEnabled = NumericEditor.Enabled;
-			NumericEditor.AccessibilityTitle = ViewModel.Property.Name + " Numeric Editor"; // TODO Localization
+			NumericEditor.AccessibilityTitle = ViewModel.Property.Name + string.Format (" {0} {1}", LocalizationResources.AccessibilityNumeric, LocalizationResources.AccessibilityEditor);
```
alanmcgovern: We should translate whole phrases, not individual words. "Numeric Editor" is the translatable phrase here!

Xamarin.PropertyEditing.Mac/Controls/CGRectEditorControl.cs
```diff
@@ -8,24 +8,23 @@ internal class CGRectEditorControl
 	{
 		public CGRectEditorControl ()
 		{
-			// TODO localize
 			XLabel.Frame = new CGRect (26, 23, 25, 24);
-			XLabel.StringValue = "X:";
+			XLabel.StringValue = LocalizationResources.XLabel + ":";
```
alanmcgovern: `X` is not translatable in this context. It refers to API. `X, Y, Width, Height` might all be better off not translated as they directly refer to actual C# properties.

Xamarin.PropertyEditing.Mac/Controls/BaseNumericEditorControl.cs
CartBlanche: fixed

Xamarin.PropertyEditing.Mac/Controls/CGRectEditorControl.cs
CartBlanche: Is this a major issue if they are already in the resx? Meaning should I pull these out of the resx?

ermau: You should pull them out of the resx.

alanmcgovern: There is a mechanism for marking words as "do not translate". This is useful for sentences where, for example, we want to put the Product name in there. Or if we refer to API like watchOS "Complications" and don't want that specific word translated, but do want it in the correct part of the translated sentence

CartBlanche: please pull in this PR if latest changes are ok.

Xamarin.PropertyEditing.Mac/Controls/BaseNumericEditorControl.cs
```diff
@@ -72,7 +72,7 @@ protected override void SetEnabled ()
 		protected override void UpdateAccessibilityValues ()
 		{
 			NumericEditor.AccessibilityEnabled = NumericEditor.Enabled;
-			NumericEditor.AccessibilityTitle = ViewModel.Property.Name + " Numeric Editor"; // TODO Localization
+			NumericEditor.AccessibilityTitle = string.Format (" {0} {1}", ViewModel.Property.Name, LocalizationResources.AccessibilityNumeric);
```
alanmcgovern: The localizable string should be `{0} Numeric Editor`and we should be calling the api like: `Xamarin.PropertyEditing.Mac.Strings.AccessibilityNumeric (ViewModel.Property.Name)`. If we reference this nuget ( https://www.nuget.org/packages/netfx-System.StringResources/ ) and set the resx files up like we did in all of the csproj files in the designer repository it will autogenerate the right c# API for you.

We then need to add a comment explaining what `{0}` is for and what an example usage of it would look like. For example `Corner Radius Numeric Editor`.

Xamarin.PropertyEditing.Mac/LocalizationResources.resx
```diff
@@ -0,0 +1,67 @@
...
+		<value>Width Editor</value>
+		<comment>Editor for Width Value</comment>
+	</data>
+	<data name="AccessibilityHeightEditor" xml:space="preserve">
+		<value>Height Editor</value>
+		<comment>Editor for Height Value</comment>
+	</data> 
+	<data name="ArrangeBy" xml:space="preserve">
+		<value>Arrange By:</value>
+		<comment>Editor for Height Value</comment>
```
alanmcgovern: Copy/pasta!

Xamarin.PropertyEditing.Mac/Controls/BaseNumericEditorControl.cs
CartBlanche: Ah ok, thanks for the clarification.

CartBlanche:
@alanmcgovern I couldn't use the latest version of System.StringResources as it breaks the build here. Used 3.0.14, which my local md-addins is also using. And that builds and appears to work correctly.

alanmcgovern:
💯

PR #67, opened 2017-10-20
[Core] Fix AsyncWorkQueue

ermau:
This fixes an issue (demonstrated in included test) where a
child/sibling to another requester is added when that requester is not
the active. The logic that was present in the reqest to set related
requester workers to active was not present when the queue was advanced.

This also significantly reduces the work the request work method does by
simply tracking the active requester. The original comment about small
queues turned out to be false and it is likely to see queues as long as
2x the number of properties.

### DIFF
--- a/Xamarin.PropertyEditing.Tests/AsyncWorkQueueTests.cs
+++ b/Xamarin.PropertyEditing.Tests/AsyncWorkQueueTests.cs
@@ -47,6 +47,25 @@ namespace Xamarin.PropertyEditing.Tests
 		}
 
 		[Test, Timeout (1000)]
+		[Description ("Worker siblings added while their requester isn't active should ALL unlock when their requester becomes active")]
+		public async Task SameRequesterAfterNonSameIsFreed ()
+		{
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing.Tests/AsyncWorkQueueTests.cs
```diff
@@ -47,6 +47,24 @@ public void Setup ()
 		}
 
 		[Test, Timeout (1000)]
+		public async Task SameRequesterAfterNonSameIsFreed ()
```
bleroy: It's unclear to me what this is testing, and how it would fail.

ermau: The expanded commit message goes into the details, this test fails on the previous implementation. The trick is that the work queue allows items that are from the same requester as the active item to immediately begin executing. The prior implementation only had this pass-through working when it was _already_ the active item, but not when it became it _later_.

PR #69, opened 2017-10-23
[Core] Fix double completion set

ermau:


bleroy:
Is it worth writing a test for that?

ermau:
@bleroy It's extremely difficult to reproduce, the existing integration with it and property view model covers it and will now properly fail with the sync context I added.

PR #74, opened 2017-11-08
Initial Common Ratio implementation

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -254,14 +289,22 @@ protected double CoerceValue (double val)
 
 		public void IncrementNumericValue ()
 		{
-			SetValue (stepper.DoubleValue + IncrementValue);
-			OnStepperActivated (stepper, EventArgs.Empty);
+			if (AllowRatios) {
+				// TODO find elegant solution to incrementing values based on cursor position
```
ermau: Shouldn't we address this before we merge?

CartBlanche:
@ermau When you have a mo, please look over this PR.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -227,19 +244,25 @@ protected void OnEditingEnded (object sender, EventArgs e)
...
 
 		void SetValue (string value)
 		{
-			//Regulates maximun and minium out of range
-			stepper.DoubleValue = CoerceValue (FieldValidation.FixInitialValue (value, Value.ToEditorString ()).ToEditorDouble ());
-			numericEditor.StringValue = FieldValidation.RoundDoubleValue (stepper.DoubleValue.ToEditorString (), NumericMode == ValidationType.Decimal ? FieldValidation.DefaultXcodeMaxRoundDigits : 0);
+			if (AllowRatios) {
+				// We should have a valid ration at this point.
+				var parts = FieldValidation.GetRatioParts (value);
+				SetCommonRatio (FieldValidation.ParseDoubleValue(parts[0]), FieldValidation.ParseDoubleValue (parts[1]), FieldValidation.GetSeparator(value));
```
ermau: As much of this behavior as possible was supposed to move to the view model, so much so that the `FieldValidation` class could be deleted entirely. This crashes when you just select all and enter a whole value.

CartBlanche: Oh damn. I will look into this!

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
```diff
@@ -0,0 +1,169 @@
...
+
+		protected override void OnValueChanged ()
+		{
+			base.OnValueChanged ();
+			OnPropertyChanged (nameof (ValueString));
+		}
+
+		void SetRatioFromString (string value)
+		{
+			var strippedValue = value.Replace (" ", string.Empty);
```
bleroy: I think you could avoid both the `Replace` and the loop by splitting once on the set of admissible separators and space, with the option to remove empty entries on.

bleroy: We could probably use a bunch of tests on this parsing logic.

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
```diff
@@ -0,0 +1,169 @@
...
+					
+			//Checks if needs to be positive value
+			if (value < 0)
+				return false;
+					
+			//Checks a common validation
+			return CommonValidate (stringValue);
+		}
+
+		public static bool CommonValidate (string finalString)
```
bleroy: Maybe make this method name explicitly state usage. This is pretty vague.

Xamarin.PropertyEditing.Mac/Controls/Custom/RatioEditor.cs
```diff
@@ -0,0 +1,78 @@
...
+		{
+			caretLocation = 0;
+			selectionLength = 0;
+			if (NumericEditor.CurrentEditor != null) {
+				caretLocation = NumericEditor.CurrentEditor.SelectedRange.Location;
+				selectionLength = NumericEditor.CurrentEditor.SelectedRange.Length;
+			}
+		}
+
+		public class RatioEventArgs
```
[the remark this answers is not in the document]
CartBlanche: It was meant to derive from EventArgs, lost in a refactor.

Xamarin.PropertyEditing/Drawing/CommonRatio.cs
```diff
@@ -0,0 +1,73 @@
...
+
+		// Normal values like 1, 200, 400 would be stored as:
+		// Numerator = 100, Denominator = 1
+		//
+		// Ratio values such as 1:3 or 5/6 would be stored as
+		// Numerator = 1, Denominator = 3 etc.
+		public double Numerator { get; set; }
+		public double Denominator { get; set; }
+
+		public double Value => Numerator / Denominator;
```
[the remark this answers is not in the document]
CartBlanche: It is handled in the VM in ValueChanged

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
CartBlanche: Agreed.

Xamarin.PropertyEditing/Drawing/CommonRatio.cs
```diff
@@ -0,0 +1,73 @@
...
+		{
+			return Equals (left, right);
+		}
+
+		public static bool operator != (CommonRatio left, CommonRatio right)
+		{
+			return !Equals (left, right);
+		}
+
+		public override bool Equals (object obj)
```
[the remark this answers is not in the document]
CartBlanche: Nicely spotted, I had overlooked that.

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
```diff
@@ -0,0 +1,167 @@
...
+					Value = new CommonRatio (numerator, 1, GetSeparator (strippedValue));
+				}
+			}
+
+			if (!parsed) {
+				SetError (string.Format (LocalizationResources.InvalidRatio, value));
+			}
+		}
+
+		char GetSeparator (string value)
```
bleroy: I think here and in `GetSeparatorPosition`, you can use `IndexOfAny`.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
```diff
@@ -54,32 +54,78 @@ public override bool ShouldBeginEditing (NSText fieldEditor)
...
+			//Checks if needs to be possitive value
+			if (!allowNegativeValues && value < 0)
+				return false;
+			//Checks a common validation
+			return ViewModels.RatioViewModel.IsValidInput (finalString);
+		}
+
+		public static bool CheckIfRatio (string finalString, ValidationType mode, bool allowNegativeValues)
+		{
+			var parts = finalString.Split (ViewModels.RatioViewModel.Separators);
```
bleroy: I think the handling of spaces got lost in the new implementation. You may need a second char array that also has spaces. Also, I think you should use the overload that tells `Split` to remove empty entries (`StringSplitOptions. RemoveEmptyEntries`).

Xamarin.PropertyEditing.Tests/RatioViewModelTests.cs
```diff
@@ -0,0 +1,182 @@
...
+		}
+
+		[Test]
+		public void RatioWithNumeratorDenominatorAndSlashSeparator ()
+		{
+			var property = GetPropertyMock ();
+			var editor = GetBasicEditor (property.Object);
+			var vm = GetViewModel (property.Object, new[] { editor });
+
+			vm.SetRatioFromString ("21/2");
```
bleroy: I'd add a test case with spaces, to finish addressing my above comment.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
CartBlanche: This is a (hopefully quick) high level check, done at the control level. The full check is done in the VM here -  https://github.com/xamarin/Xamarin.PropertyEditing/pull/74/files/9257d48169ec83a77d98286fd094fa5c73aa5ee7#diff-995a4c01ec15e058d67645b539295624R102

Xamarin.PropertyEditing.Tests/RatioViewModelTests.cs
CartBlanche: Yes, good idea.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
bleroy: Ah, OK, I did miss that, but in fact, I think you can avoid the `Replace` altogether by including space in the split chars.

CartBlanche: I did think about adding a 2nd array to include the space, but I thought the code would be simpler to maintain without needing 2 arrays, or if having one 3 element array we'd need to generate a temporary 2 element array when checking for valid separators. 
Is your concern the speed overhead of the `Replace` call?

bleroy: No, I don't think speed is a concern on this code path. I'm mostly nitpicking at this point. It's your call really, but having the second array right next to the first one didn't seem necessarily like it would cause any big maintenance problems.

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
CartBlanche: I always forget about IndexOfAny.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
CartBlanche: OK, You've convinced me :)

ermau:
- [ ] The value can start at `0:0`, but I can't go back to `0:0`
- [ ] Selecting any part acts as the whole thing is selected for incrementing/decrementing
- [ ] Don't we have icons for the up/down buttons? They're just blank and don't match the designs.
- [ ] Hitting tab just goes up to the property filter.

Xamarin.PropertyEditing.Tests/ObservableLookupTests.cs
```diff
@@ -56,7 +56,7 @@ public void RemoveLastItemInGroup (string key)
 			var grouping = lookup[key];
 			((INotifyCollectionChanged) grouping).CollectionChanged += (sender, args) => {
 				if (args.Action == NotifyCollectionChangedAction.Remove) {
-					if (args.OldItems[0] == value)
+					if ((string)args.OldItems[0] == value)
```
ermau: Why?

CartBlanche: To remove compiler warnings like `Warning CS0252: Possible unintended reference comparison; to get a value comparison, cast the left hand side to type 'string' (CS0252) (Xamarin.PropertyEditing.Tests)`

CartBlanche: Btw, there are other warnings in the unit tests. If you'd rather we don't address them, I'll revert this change.

CartBlanche:
@ermau I've already discussed icons with @vancura about the spinner. Plan was to handle that when we do the UI pretty pass.

CartBlanche:
@vancura 's new images are now part of this PR. Looks pretty snazzy!

CartBlanche:
Conflicts resolved!

ermau:
When you select the entire contents and keep incrementing until the denominator gains a second digit, the second digit is not selecting breaking the selection of the whole contents and starts incrementing only the numerator.

 - [ ] We should maintain full selection
 - [ ] If the selection goes over one number into the other, we should either increment both still, or not increment either.

Xamarin.PropertyEditing.Mac/Controls/Custom/NumericSpinEditor.cs
```diff
@@ -174,106 +205,106 @@ public NumericSpinEditor ()
...
-			//Regulates maximun and minium out of range
-			if (!string.IsNullOrEmpty (value)) {
-				stepper.DoubleValue = CoerceValue (FieldValidation.FixInitialValue (value, Value.ToEditorString ()).ToEditorDouble ());
-				numericEditor.StringValue = FieldValidation.RoundDoubleValue (stepper.DoubleValue.ToEditorString (), NumericMode == ValidationType.Decimal ? FieldValidation.DefaultXcodeMaxRoundDigits : 0);
-			} else {
-				stepper.StringValue = value;
-				numericEditor.StringValue = value;
-			}
+			numericEditor.StringValue = value;
+			ValueChanged?.Invoke (this, EventArgs.Empty);
```
lewing: This appears to invoke ValueChanged even if the value hasn't changed?

CartBlanche: @lewing Do you mean a negative ratios arn't working? If you mean normal float/integer that appears to work for me, as per this screenshot.
screen shot 2018-05-17 at 14 13 49

alanmcgovern: Does he mean values like `0.5`?

Also, did you mean to include a screenshot or a screencast? :p I do love screencasts :)

lewing: yeah, I was using the NumericSpinEditor to handle values that must be between 0 and 1, the NSStepper version handled that as expected by limiting to Min/Max and incrementing by the Step increment of .01 when clicking up/down.

CartBlanche: thanks @lewing I'll double check the original behaviour.

CartBlanche: @lewing something like this - https://www.screencast.com/t/PenH2m0lU0L ?

alanmcgovern:
I have a proposal!

When the *denominator* is selected (and nothing else!) then the increment arrows will affect the denominator only.

When the selection is anything other than *just* the denominator, we increment only the numerator.

ermau:
> The most common case is a value of 1 for constraints in iOS. Incrementing this is pretty easy.

Do they show 1 or 1:1? Would it make more sense to start as a whole number?

> When the selection is anything other than just the denominator, we increment only the numerator.

This seems more confusing to me., if I select the whole thing, why wouldn't everything increment?

alanmcgovern:
They show `1` by default and increment by `1` by default.

CartBlanche:
@alanmcgovern So if we start at 1:1, that's what iOS and other users would logically expect. 
I also think if you select the whole thing it should increment both and if nothing is selected then only the numerator increments. 
I think we should try to provide expected behaviour (like the numerator increment), but also make it a better UX where possible.

CartBlanche:
Done!

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
```diff
@@ -0,0 +1,153 @@
...
+			set {
+				if (Value.StringValue.Equals (value))
+					return;
+
+				SetRatioFromString (value);
+			}
+		}
+
+		public static char[] Separators = { ':', '/' };
+		public static char[] SplitSeparators = { ':', '/', ' ' };
```
ermau: I don't think that we want to accept spaces.

CartBlanche:
@ermau All the numeric controls behave that way and from trying it on xcode that is how xcode behaves. When the error beep sounds it doesn't change focus until the error is fixed. I don't think auto-magically reseting the value and disappearing to another control is a good UX. Reset the value, but by staying in the same place the user knows that something wasn't right. Otherwise they tab thinking everything is fine and move onto the next control, only to find out later on that at slight typo reset their value. If it stays in place they know straight away it was wrong.

Xamarin.PropertyEditing.Tests/RatioViewModelTests.cs
```diff
@@ -0,0 +1,119 @@
...
+	internal class RatioViewModelTests
+		: PropertyViewModelTests<CommonRatio, RatioViewModel>
+	{
+		[Test]
+		public void Numerator ()
+		{
+			var property = GetPropertyMock ();
+			var editor = GetBasicEditor (property.Object);
+			var vm = GetViewModel (property.Object, new[] { editor });
+			Assume.That (vm.Value, Is.EqualTo (new CommonRatio (3, 1, ':')));
```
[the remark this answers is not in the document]
CartBlanche: They were just arbitrary values. Just to test the value had changed. An empty constructor which defaults to 1, 1, : would work just as well.  I'll make it that.

Xamarin.PropertyEditing/Drawing/CommonRatio.cs
```diff
@@ -0,0 +1,105 @@
...
+	public struct CommonRatio : IEquatable<CommonRatio>
+	{
+		/// <param name="numerator">The ratio numerator value.</param>
+		/// <param name="denominator">The ratio denominator value.</param>
+		/// <param name="ratioSeparator">The ratio separator character, '/' or ':'.</param>
+		public CommonRatio (double numerator, double denominator, char ratioSeparator)
+		{
+			this.numerator = numerator;
+			this.denominator = denominator;
+			this.ratioSeparator = ratioSeparator;
```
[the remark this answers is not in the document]
CartBlanche: Fixed.

Xamarin.PropertyEditing/ViewModels/RatioViewModel.cs
CartBlanche: This was to more easily split should the user enter something like " 21: 2 ". In theory the `double.TryParse()` check should handle it. But if you feel strongly about it, I can remove it.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -33,108 +34,173 @@ public PropertyButton ()
...
+		private void ToggleFocusImage (bool focused = false)
 		{
-			switch (valueSource) {
+			if (viewModel != null) {
+				
+				switch (viewModel.ValueSource) {
 				case ValueSource.Binding:
-					Image = activeImage ? NSImage.ImageNamed ("property-button-bound-mac-active-10") : NSImage.ImageNamed ("property-button-bound-mac-10");
+					switch (PropertyEditorPanel.ThemeManager.Theme) {
+					case PropertyEditorTheme.Dark:
```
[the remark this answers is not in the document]
CartBlanche: Yes.

Xamarin.PropertyEditing.Tests/PanelViewModelTests.cs
```diff
@@ -18,7 +18,7 @@ internal class PanelViewModelTests
 		: PropertiesViewModelTests<PanelViewModel>
 	{
 		[SetUp]
-		public void Setup ()
+		public override void Setup ()
```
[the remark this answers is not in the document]
CartBlanche: Fixed.

CartBlanche:
@ermau I believe I've responded, either code-wise or text-wise, to your requested changes.

Xamarin.PropertyEditing/Drawing/CommonRatio.cs
```diff
@@ -0,0 +1,105 @@
...
+				return denominator;
+			}
+			set { denominator = value; }
+		}
+
+		public double Value => Numerator / Denominator;
+
+		// This can be either '/' or ':'
+		char ratioSeparator;
+		public char RatioSeparator {
```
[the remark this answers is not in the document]
CartBlanche: Possibly so, but in the case of CommonRatio it is also input data, not pure formatting/presentation. Unlike say entering 1000 and a field gets auto-formatted to 1,000 or 1.000, depending on region. When @alanmcgovern and I discussed it last year we thought having it part of the struct made sense and didn't add any unnecessary overhead. I'll let you 2 discuss whether it should be in there or not.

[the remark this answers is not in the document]
ermau: How is it input data? What's the functional difference between `/` and `:` ?

PR #75, opened 2017-11-14
[Win] Group editor

ermau:


Xamarin.PropertyEditing/ViewModels/PropertyGroupViewModel.cs
```diff
@@ -0,0 +1,103 @@
...
+				if (vm.IsAvailable)
+					this.propertiesView.Add (vm);
+
+				vm.PropertyChanged += OnChildPropertyChanged;
+			}
+		}
+
+		public IReadOnlyList<PropertyViewModel> Properties => this.propertiesView;
+
+		public override string Name
```
bleroy: Shouldn't this be initialized or have an actual getter?

Xamarin.PropertyEditing/ViewModels/PropertyGroupViewModel.cs
```diff
@@ -0,0 +1,103 @@
...
+				for (int i = 0; i < current.Count; i++) {
+					var vm = current[i];
+					if (!MatchesFilter (vm))
+						this.propertiesView.Remove (vm);
+				}
+			} else {
+				this.propertiesView.Reset (this.properties.Where (MatchesFilter).OrderBy (p => p.Property.Name));
+			}
+
+			OnPropertyChanged (nameof(HasChildElements));
```
bleroy: Won't this fire even if you go from a filter that yields zero children to another that also yields none, and vice-versa?

Xamarin.PropertyEditing.Windows/GroupEditorControl.cs
```diff
@@ -0,0 +1,116 @@
...
+		protected override void OnSelectionChanged (SelectionChangedEventArgs e)
+		{
+			if (e.AddedItems.Count == 0)
+				return;
+
+			var presenter = (ContentPresenter) ItemContainerGenerator.ContainerFromItem (e.AddedItems[0]);
+			if (presenter == null || VisualTreeHelper.GetChildrenCount (presenter) == 0)
+				return;
+
+			var toggle = (ToggleButton) VisualTreeHelper.GetChild (presenter, 0);
```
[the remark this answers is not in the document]
ermau: We are, yes. Going down to `ToggleButton` instead of a radio is about the best we can do. Any deviation from that means we've broken the control, so I'd rather it explodes.

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
```diff
@@ -79,27 +81,69 @@ public void SetIsExpanded (string group, bool isExpanded)
...
+
+				if (remainingItems != null)
+					this.arranged.Add (grouping.Key, remainingItems);
+				else
+					this.arranged.Add (grouping);
+			}
+
+			if (groupedTypeProperties != null) {
+				int i = 0;
+				foreach (var kvp in groupedTypeProperties.OrderBy (kvp => kvp.Key)) {
```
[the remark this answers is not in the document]
ermau: They're unsorted at this point, `Dictionary` does not maintain order. I could potentially switch it to `ObservableLookup` which will maintain order and have a hash key lookup, but it's a lot heavier.

Xamarin.PropertyEditing.Windows/GroupEditorControl.cs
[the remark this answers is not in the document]
bleroy: Right, but in later parts of the code, you're detecting that and throwing a nicer exception, so might as well do it here too...

Xamarin.PropertyEditing/ViewModels/PropertyGroupViewModel.cs
ermau: No, property groups don't have names.

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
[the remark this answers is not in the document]
bleroy: You're right, `arranged` is sorted at all times, but groupedTypeProperties is not.

Xamarin.PropertyEditing/ViewModels/PropertyGroupViewModel.cs
bleroy: I'd be explicit and return null then.

Xamarin.PropertyEditing/ViewModels/PropertyGroupViewModel.cs
ermau: Technically, but it was the less error prone route with minimal impact. I'll see about changing it.

PR #80, opened 2017-11-21
Fix null ref when getting ValueSource if value is null

bleroy:
This was happening specifically when switching fast enough from one inspected mock editable object to another in the Windows standalone app.


Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -23,7 +23,7 @@ public PropertyViewModel (IPropertyInfo property, IEnumerable<IObjectEditor> edi
 			UpdateCurrentValue ();
 		}
 
-		public ValueSource ValueSource => this.value.Source;
```
ermau: `ValueSource` does not combine, this should just be `ValueSource.Default`. Also let's add a test that this property is `Default` when there's no value source (like when two editors are present and giving different values.)

PR #82, opened 2017-11-23
Bleroy brush editor 2

bleroy:
Solid brush editor moved to a tab in preparation for future color resource tab. Also put a property presenter around, and did the same to the color space drop-down.

Note: the view model is not 100% ready for bindings and resources on deep properties such as the opacity, color space, and solid brush color. This is known, so I added TODOs in there to that effect.

### DIFF
--- a/Xamarin.PropertyEditing.Windows.Standalone/MockedSampleControlButton.cs
+++ b/Xamarin.PropertyEditing.Windows.Standalone/MockedSampleControlButton.cs
@@ -23,17 +23,23 @@ namespace Xamarin.PropertyEditing.Windows.Standalone
 			MockedControl.AddProperty<CommonBrush> (this.readOnlyBrushPropertyInfo);
 		}
 
-		public async Task SetBrush (IObjectEditor editor, CommonBrush brush)
+		public async Task SetBrushInitialValue (IObjectEditor editor, CommonBrush brush)
 		{
+			if (this.brushSet) return;
 			await editor.SetValueAsync (this.brushPropertyInfo, new ValueInfo<CommonBrush> { Value = brush });
+			this.brushSet = true;
 		}
 
-		public async Task SetReadOnlyBrush (IObjectEditor editor, CommonBrush brush)
+		public async Task SetReadOnlyBrushInitialValue (IObjectEditor editor, CommonBrush brush)
 		{
+			if (this.readOnlyBrushSet) return;
 			await editor.SetValueAsync (this.readOnlyBrushPropertyInfo, new ValueInfo<CommonBrush> { Value = brush });
+			this.readOnlyBrushSet = true;
 		}
 
-		private IPropertyInfo brushPropertyInfo;
-		private IPropertyInfo readOnlyBrushPropertyInfo;
+		IPropertyInfo brushPropertyInfo;
+		IPropertyInfo readOnlyBrushPropertyInfo;
+		bool brushSet = false;
+		bool readOnlyBrushSet = false;
 	}
 }
--- a/Xamarin.PropertyEditing.Windows.Standalone/MockedSampleControlButton.cs
+++ b/Xamarin.PropertyEditing.Windows.Standalone/MockedSampleControlButton.cs
@@ -23,23 +23,23 @@ namespace Xamarin.PropertyEditing.Windows.Standalone
 			MockedControl.AddProperty<CommonBrush> (this.readOnlyBrushPropertyInfo);
 		}
 
-		public async Task SetBrushInitialValue (IObjectEditor editor, CommonBrush brush)
+		public async Task SetBrushInitialValueAsync (IObjectEditor editor, CommonBrush brush)
 		{
 			if (this.brushSet) return;
 			await editor.SetValueAsync (this.brushPropertyInfo, new ValueInfo<CommonBrush> { Value = brush });
 			this.brushSet = true;
 		}
 
-		public async Task SetReadOnlyBrushInitialValue (IObjectEditor editor, CommonBrush brush)
+		public async Task SetReadOnlyBrushInitialValueAsync (IObjectEditor editor, CommonBrush brush)
 		{
 			if (this.readOnlyBrushSet) return;
 			await editor.SetValueAsync (this.readOnlyBrushPropertyInfo, new ValueInfo<CommonBrush> { Value = brush });
 			this.readOnlyBrushSet = true;
 		}
 
-		IPropertyInfo brushPropertyInfo;
-		IPropertyInfo readOnlyBrushPropertyInfo;
-		bool brushSet = false;
-		bool readOnlyBrushSet = false;
+		private IPropertyInfo brushPropertyInfo;
+		private IPropertyInfo readOnlyBrushPropertyInfo;
+		private bool brushSet = false;
+		private bool readOnlyBrushSet = false;
 	}
 }
--- a/Xamarin.PropertyEditing.Windows/BrushEditorControl.cs
+++ b/Xamarin.PropertyEditing.Windows/BrushEditorControl.cs
@@ -1,5 +1,5 @@
 using System.Linq;
-using System.Windows.Controls;
+using System.Windows;
 using System.Windows.Controls.Primitives;
 using System.Windows.Input;
 using Xamarin.PropertyEditing.ViewModels;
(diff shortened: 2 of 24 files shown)


ermau:
The hex editor text box doesn't release it's full selection after you click a second time. See `StringEditorControl` for an implementation of fully select on first select (maybe we should break that out into a TextBox subclass for reuse).

ermau:
Also it doesn't appear the property button for the editor is hooked up.

ermau:
Bugs:
 - Alpha value starts incorrect at 0%, but true value is 100%.
 - Resetting the value does not appear to work correctly. Wrong brush type is still selected, alpha values don't match.

ermau:
More issues:
 - [ ] RGBA editors are unthemed
 - [ ] RGBA editors don't accept enter to update value
 - [ ] Hue slider will not adjust until color has been changed
 - [ ] Editor tab seems to be missing?
 - [ ] Property button next to color space seems to actually control the brush value
   - [ ] If it is for the brush value, resetting clears the brush value but the type tab is not set back to "no brush"

bleroy:
So for the property button next to color space (and opacity), that's expected for now, until the subproperty story is in place. It's pretty much a placeholder.

ermau:
- [ ] Resetting from the editor tab button resets the value, but the value change doesn't put the type selector back to no brush. (Reworded description of last issue from above).
- [ ] Resetting and reopening has value source changed, but value hasn't changed.
- [ ] After doing ^, selecting solid results in a mismatch between value and picker.

bleroy:
I don't think I should address the editor tab button issue, at least not now: that button is a placeholder for now, it's not supposed to reset the brush as a whole, only the Color property of the solid brush, so it shouldn't put the brush type selector to no brush.

I'm fixing the other two.

Xamarin.PropertyEditing.Windows/BrushEditorControl.cs
```diff
@@ -22,9 +22,12 @@ public override void OnApplyTemplate ()
 			this.brushTabs = this.brushBoxPopup?.Child?.GetDescendants<BrushTabbedEditorControl>().FirstOrDefault();
 
 			if (IsEnabled && this.brushBoxButton != null && this.brushBoxPopup != null) {
```
ermau: `IsEnabled` can change after the template is applied.

bleroy: Yes, actually, we don't even need that test, as the button that triggers the popup is already bound to `IsEnabled` in the XAML.

ermau:
- [ ] Last color should be the last "resting" color, instead it just matches whatever's currently being dragged around.
- [ ] Initial color stays the color the property was initially in its lifetime, rather than the initial color from when the editor was opened.
- [ ] Hue in HLS walks:
 - [ ] Dragging off the palette results in Hue going to 0

Xamarin.PropertyEditing.Windows/ColorComponentBox.cs
```diff
@@ -56,25 +57,31 @@ public override void OnApplyTemplate ()
 		{
 			base.OnApplyTemplate ();
 
-			this.innerTextBox = GetTemplateChild ("innerTextBox") as TextBox;
+			this.innerTextBox = GetTemplateChild ("innerTextBox") as TextBoxEx;
 			if (this.innerTextBox != null) {
```
ermau: The control is invalid without this, it should throw if the part isn't present.

Xamarin.PropertyEditing/ViewModels/BrushPropertyViewModel.cs
```diff
@@ -47,7 +48,11 @@ public BrushPropertyViewModel (TargetPlatform platform, IPropertyInfo property,
 		protected override async Task UpdateCurrentValueAsync ()
 		{
 			await base.UpdateCurrentValueAsync ();
+			if (Solid != null) {
+				Solid.Reset();
+			}
 			OnPropertyChanged (nameof (Opacity));
+			OnPropertyChanged (nameof (Solid));
```
ermau: This is incorrect, `Solid` has not changed, it stays the same value from `ctor`. This will force all `Solid.*` bindings to re-query after they've already requeried the individual properties that update from `Reset`.

Xamarin.PropertyEditing/ViewModels/BrushPropertyViewModel.cs
```diff
@@ -47,7 +48,11 @@ public BrushPropertyViewModel (TargetPlatform platform, IPropertyInfo property,
 		protected override async Task UpdateCurrentValueAsync ()
```
ermau: This isn't the right override for these property change notifications. A request to update the current value does not necessarily mean the value changes. `OnValueChanged()` is the proper override.

Xamarin.PropertyEditing/ViewModels/SolidBrushViewModel.cs
```diff
@@ -92,6 +100,17 @@ public void CommitShade ()
 			Parent.Value = new CommonSolidBrush (Shade, ColorSpace, opacity);
 		}
 
+		public void Reset()
+		{
+			this.hueColor = null;
+			this.shade = null;
+			this.lastColor = null;
+			OnPropertyChanged (nameof (Color));
```
ermau: This is duplicated. When `Reset()` is called from your `UpdateCurrentValue` override, it's signaling a property changed notice for `Value`, which you've already hooked to update this property (but not `LastColor` or `Shade` oddly, and `ColorSpace` is missing from here).

Xamarin.PropertyEditing.Windows/TextBoxEx.cs
```diff
@@ -116,12 +116,14 @@ private void FocusSelect()
...
 
+		internal bool BubbleAfterReturn { get; set; } = false;
+
 		private void OnPreviewKeyDown (object sender, System.Windows.Input.KeyEventArgs e)
 		{
 			if (e.Key == Key.Enter) {
 				var expression = GetBindingExpression (TextProperty);
 				expression?.UpdateSource ();
-				e.Handled = true;
+				if (!BubbleAfterReturn) e.Handled = true;
```
ermau: We can actually probably just remove the handled setter altogether, I don't think it's needed.

Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
```diff
@@ -47,6 +47,15 @@ public bool IsSubProperty
 			private set { SetValue (IsSubPropertyPropertyKey, value); }
 		}
 
+		private static readonly DependencyProperty ShowPropertyButtonProperty = DependencyProperty.Register (
```
ermau: Why is this private?

Xamarin.PropertyEditing/ViewModels/SolidBrushViewModel.cs
bleroy: I since got rid of this.

bleroy:
[ ] The behavior when a brush property is set to a resource is not correct.

Xamarin.PropertyEditing.Windows.Standalone/MockedSampleControlButton.cs
```diff
@@ -23,17 +23,23 @@ public MockedSampleControlButton () : base (new MockSampleControl ())
 			MockedControl.AddProperty<CommonBrush> (this.readOnlyBrushPropertyInfo);
 		}
 
-		public async Task SetBrush (IObjectEditor editor, CommonBrush brush)
+		public async Task SetBrushInitialValue (IObjectEditor editor, CommonBrush brush)
```
ermau: Returns `Task`? `Async` postfixed.

Xamarin.PropertyEditing.Windows.Standalone/MockedSampleControlButton.cs
```diff
@@ -23,17 +23,23 @@ public MockedSampleControlButton () : base (new MockSampleControl ())
...
+		public async Task SetReadOnlyBrushInitialValue (IObjectEditor editor, CommonBrush brush)
 		{
+			if (this.readOnlyBrushSet) return;
 			await editor.SetValueAsync (this.readOnlyBrushPropertyInfo, new ValueInfo<CommonBrush> { Value = brush });
+			this.readOnlyBrushSet = true;
 		}
 
-		private IPropertyInfo brushPropertyInfo;
-		private IPropertyInfo readOnlyBrushPropertyInfo;
+		IPropertyInfo brushPropertyInfo;
```
ermau: Don't remove explicit visibility when it's preferred.

Xamarin.PropertyEditing/ViewModels/SolidBrushViewModel.cs
```diff
@@ -18,49 +18,47 @@ public SolidBrushViewModel (BrushPropertyViewModel parent, IReadOnlyList<string>
...
 				if (!this.hueColor.Equals (value)) {
-					var saturation = Color.Saturation;
-					var brightness = Color.Brightness;
-					Color = CommonColor.FromHSB (value.Hue, saturation, brightness, Color.A);
 					this.hueColor = value;
 					OnPropertyChanged ();
-					Parent.Value = new CommonSolidBrush (Color, ColorSpace, Parent.Value.Opacity);
+
+					var saturation = Shade.Saturation;
+					// We should not update the color value on a hue change if the current color is unsaturated, as grey has no hue.
```
ermau: Test?

bleroy: I added a few.

PR #83, opened 2017-11-24
UI Tweaks to standardise RowHeight, ControlHeight, Font and Control Alignments

CartBlanche:
What the current tweaks look like...
screen shot 2017-12-05 at 12 30 58

screen shot 2017-12-05 at 12 31 14


Xamarin.PropertyEditing.Mac/Controls/BaseEditorControl.cs
```diff
@@ -67,9 +105,22 @@ protected void SetErrors (IEnumerable errors)
...
+		{
+			PropertyButtonClicked?.Invoke (this, EventArgs.Empty);
+		}
+
+		void NotifyActioButtonClicked ()
+		{
+			ActionButtonClicked?.Invoke (this, EventArgs.Empty);
+		}
+
+		protected abstract void RestDefaultValues ();
```
ermau: What is `RestDefaultValues` ?

CartBlanche: @ermau This was meant to hook into the property event when the user decides to reset to that property's default value. I will remove this as the event is there, but would like to talk to you about the intricacies of the implementation for the various states.

PR #85, opened 2017-12-01
Add CommonRectangle Editors. Needed in UITweaks.

CartBlanche:


Xamarin.PropertyEditing/ViewModels/RectanglePropertyViewModel.cs
```diff
@@ -1,14 +1,63 @@
...
+			get { return Value.Height; }
+			set {
+				if (Value.Height == value)
+					return;
+
+				Value = new CommonRectangle (Value.X, Value.Y, Value.Width, value);
+			}
+		}
+
+		protected override void OnValueChanged ()
```
bleroy: It looks like setting any of the four properties will trigger a change notification on all of them, through the setting of Value. On the other hand, we're not being more subtle on `Size`, so this is probably fine for now.

CartBlanche: @bleroy I didn't want to stray too much from the pattern already used in the CommonSize and CommonPoint editors where if one property was set, then both were set. Maybe go with this for now, we can always make it more granular later on, if we think it will give a better user experience?

PR #86, opened 2017-12-05
Theme improvements

ermau:
image

This PR:
 - Correctly themes combo boxes for light and dark
 - Fixes labeling for the size editor
 - Fixes the text box border highlight to match
 - Matches string editor focus behavior

### DIFF
--- a/Xamarin.PropertyEditing.Windows/StringEditorControl.cs
+++ b/Xamarin.PropertyEditing.Windows/StringEditorControl.cs
@@ -1,7 +1,43 @@
-﻿namespace Xamarin.PropertyEditing.Windows
+using System;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Input;
+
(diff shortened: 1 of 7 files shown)


Xamarin.PropertyEditing.Windows/StringEditorControl.cs
```diff
@@ -1,7 +1,43 @@
...
+				throw new InvalidOperationException ("StringEditorControl must have a 'TextBox' TextBox part");
+
+			this.textBox.PreviewMouseLeftButtonDown += OnLeftDown;
+			this.textBox.GotKeyboardFocus += OnGotFocus;
+			this.textBox.MouseDoubleClick += OnGotFocus;
+		}
+
+		private TextBox textBox;
+
+		private void OnGotFocus (object sender, EventArgs e)
```
bleroy: nit: not a fan of the handler's name here, since it's used on double click as well, and it doesn't convey what it does either.

PR #121, opened 2017-12-08
Windows category memory

### LINKED ISSUE #93: Windows category arrange memory

ermau:
This PR:
 - Closed #93, making the category expansion/closure remember and filtering auto-expand categories
 - Fixes a bug leaving empty categories in place when cleared from grouped properties
 - Fixes a crash from the group editor control metadata override.

### DIFF
--- a/Xamarin.PropertyEditing.Windows/CategoryExpander.cs
+++ b/Xamarin.PropertyEditing.Windows/CategoryExpander.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Data;
 using System.Windows.Media;
 using Xamarin.PropertyEditing.ViewModels;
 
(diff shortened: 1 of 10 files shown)


Xamarin.PropertyEditing.Windows/CategoryExpander.cs
```diff
@@ -0,0 +1,99 @@
...
+		{
+			this.fromFilter = true;
+			UpdateValue ();
+			this.fromFilter = false;
+		}
+
+		private void UpdateViewModel ()
+		{
+			FrameworkElement element = this;
+			while (element != null && !(element is ItemsControl)) {
```
bleroy: FYI, it happens I just added a helper for that: `XamlHelper.FindAncestor<T>(this DependencyObject control)`

Xamarin.PropertyEditing/CategoryComparer.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+	internal class CategoryComparer
+		: IComparer<string>
+	{
+		public static readonly CategoryComparer Instance = new CategoryComparer();
+
+		public int Compare (string x, string y)
+		{
+			int result = Comparer<string>.Default.Compare (x, y);
+			if (result != 0 && (String.IsNullOrEmpty (x) || String.IsNullOrEmpty (y)))
```
bleroy: It would be good to doc-comment how this differs from the default string comparison, which is if I understand correctly that null goes last.

Xamarin.PropertyEditing.Windows/CategoryExpander.cs
ermau: Which is in an unmerged PR...

PR #125, opened 2017-12-19
Add ThicknessPropertyViewModel with change notification

lewing:
The thickness editor wasn't actually editing the thickness
values, this adds the needed logic.

### DIFF
--- a/Xamarin.PropertyEditing.Tests/ThicknessPropertyViewModelTests.cs
+++ b/Xamarin.PropertyEditing.Tests/ThicknessPropertyViewModelTests.cs
@@ -1,0 +1,152 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Xamarin.PropertyEditing.Drawing;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Tests
+{
+	[TestFixture]
+	internal class ThicknessPropertyViewModelTests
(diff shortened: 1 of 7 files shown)


Xamarin.PropertyEditing.Tests/ThicknessPropertyViewModelTests.cs
```diff
@@ -0,0 +1,152 @@
...
+			bool topChanged = false, valueChanged = false;
+			vm.PropertyChanged += (sender, args) => {
+				if (args.PropertyName == nameof (ThicknessPropertyViewModel.Top))
+					topChanged = true;
+				if (args.PropertyName == nameof (ThicknessPropertyViewModel.Value))
+					valueChanged = true;
+			};
+
+			vm.Top = 5;
+			Assert.That (vm.Value.Top, Is.EqualTo (5));
```
bleroy: Note: notifications for all four properties, plus Value, will be triggered in all these cases. Not sure if we care.

Xamarin.PropertyEditing/Xamarin.PropertyEditing.csproj
```diff
@@ -142,8 +141,6 @@
       <LastGenOutput>LocalizationResources.Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="Themes\" />
-  </ItemGroup>
+  <ItemGroup />
```
bleroy: Not sure removing those folders was a conscious decision, but that empty `ItemGroup` can safely be removed.

PR #128, opened 2018-01-09
[loc] Update translations

alanmcgovern:


alanmcgovern:
I just ran the loc tool and it produced this.

PR #142, opened 2018-01-25
Core/Windows Custom Expressions

### LINKED ISSUE #132: Custom expression core
The ability to set an non-interpreted text value on a property if enabled. This is effectively just a way to pass through text directly to underlying XML values.

### LINKED ISSUE #133: Windows custom expression
- [x] Core - #132
 - [x] Design
 - [x] Localized
 - Behavior:
   - [x] Keyboard control
   - [x] Accessibility
   - [x] Selection
 - Theming:
   - [x] Light themed
   - [x] Dark themed
   - [ ] High contrast

ermau:
This PR adds the core components of custom expressions and the Windows experience for them:

 - Closes #132 
 - Closes #133

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
@@ -14,13 +14,6 @@ namespace Xamarin.PropertyEditing.Mac
 		internal const string PropertyListColId = "PropertiesList";
 		internal const string PropertyEditorColId = "PropertyEditors";
 
-		static PropertyEditorPanel ()
-		{
-			PropertiesViewModel.ViewModelMap.Add (typeof (CGPoint), (p, e) => new PropertyViewModel<CGPoint> (p, e));
-			PropertiesViewModel.ViewModelMap.Add (typeof (CGRect), (p, e) => new PropertyViewModel<CGRect> (p, e));
-			PropertiesViewModel.ViewModelMap.Add (typeof (CGSize), (p, e) => new PropertyViewModel<CGSize> (p, e));
-		}
-
 		public PropertyEditorPanel ()
 		{
 			Initialize ();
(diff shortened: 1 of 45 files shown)


bleroy:
A few questions:


1. Shouldn’t a value source be explicitly specified when setting a value info with a custom expression?
2. Why are you removing `CGPoint`, `CGRect`, and `CGSize` from the ViewModelMap? I didn’t see how it was related to the rest of the change.
3. I see that you set Brush as a grouped property. What is the desired behavior for this? All I see is that they disappear altogether when selecting multiple objects.
4. The behavior seems to diverge from the WPF editor in that when there is a value, WPF seems to pre-fill the custom expression with it.
5. Do we want to run the design by @vaclav on this? It’s what WPF does, but I find the lack of non-keyboard UI, such as an OK button, weird and confusing. The total lack of visual indicator once the custom expression has been set is also disconcerting.

ermau:
> Shouldn’t a value source be explicitly specified when setting a value info with a custom expression?

No. The expression might define the value source, it's up to the target platform to figure it out and come back with the proper source when we re-query the value.

> I see that you set Brush as a grouped property. What is the desired behavior for this? All I see is that they disappear altogether when selecting multiple objects.

This is unrelated, but this is what defines them as going altogether in one category with the group editor (selectable properties and single inline editor). The template selection isn't complete for this so it's not quite working yet, this is next on my list.

> Why are you removing CGPoint, CGRect, and CGSize from the ViewModelMap? I didn’t see how it was related to the rest of the change.

The delegate signature changed so they would need to be updated. We should have already moved platform specific types in favor of the common types, so now seems like the time to remove them.

>The behavior seems to diverge from the WPF editor in that when there is a value, WPF seems to pre-fill the custom expression with it.

This is up to the provider to provide, the mocks don't actually have an expression backing so in reality should have this disabled. When you set a custom expression and re-open it, the custom expression is prefilled in.

> Do we want to run the design by @vancura on this? It’s what WPF does, but I find the lack of non-keyboard UI, such as an OK button, weird and confusing. 

Given that he already designed a Mac UI that behaves the same way, I'd say it's fine as is.

> The total lack of visual indicator once the custom expression has been set is also disconcerting.

That's not quite accurate. Any expression that a target platform can evaluate into a proper value would show the change in value. This just simply isn't part of the mock providers. Entering `{StaticResource Resource}` on WPF would come back with the resource value and a resource value indicator.

PR #143, opened 2018-01-29
Hooking up PropertyButton to ResetCommand

CartBlanche:
So that we have parity functionality of Property Resets with Windows.


ermau:
This doesn't build..

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Windows.Input;
+using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public class CommandMenuItem : NSMenuItem
```
ermau: This doesn't track `ICommand.CanExecuteChanged`

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -0,0 +1,74 @@
...
+
+			Activated += (sender, e) => { 
+				var menuOrigin = this.Superview.ConvertPointToView (new CGPoint (Frame.Location.X - 1, Frame.Location.Y + Frame.Size.Height + 4), null);
+
+				var popupMenuEvent = NSEvent.MouseEvent (NSEventType.LeftMouseDown, menuOrigin, (NSEventModifierMask)0x100, 0, this.Window.WindowNumber, this.Window.GraphicsContext, 0, 1, 1);
+
+				NSMenu.PopUpContextMenu (popUpContextMenu, popupMenuEvent, this);
+			};
+
+			popUpContextMenu = new NSMenu ();
```
ermau: We could wait to construct this until we actually are opening the menu, instead of constructing one for 100 properties that won't open it.

Xamarin.PropertyEditing.Mac/Controls/StringEditorControl.cs
```diff
@@ -75,5 +75,13 @@ protected override void UpdateAccessibilityValues ()
 		{
 			StringEditor.AccessibilityTitle = string.Format (LocalizationResources.AccessibilityString, ViewModel.Property.Name);
 		}
+
+		protected override void OnPropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
+		{
+			base.OnPropertyChanged (sender, e);
+
+			if (e.PropertyName == "ValueSource")
+				PropertyButton.ValueSourceChanged (ViewModel.ValueSource);
```
ermau: We shouldn't have to hook this up for every editor. Ideally we'd pass the view model to the property button and it'd hook everything up. Barring that at least do it where you've hooked up the reset command.

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
CartBlanche: Unless I'm missing something, as we have a reference to the actual ClearValueCommand, that will fire automagically, inside ValidateMenuItem, when command.CanExecute is called and under the hood CanClearValue () is called to make sure it can. The behaviour on both Windows and Mac, from the UI perspective, is identical. They enable and disable as they are supposed to.

ermau: That only works assuming that the can execute state doesn't change _while_ the item is being displayed. Presumably, they are only calling validate when the menu opens.

CartBlanche: Yes it is only called when you try to open the menu. 
IIRC Execute, calls CanExecute again, just before actually executing the required call. At least that is how I've written ICommands in the past and gets around that very scenario.

CartBlanche: But in this case, how is the VM for that particular property likely to change, without user interaction?

ermau: Maybe there's some button we hit in the menu itself that doesn't end up closing it. Maybe some async update comes in after the menu opens.

We shouldn't cut corners just because we can't think of a scenario where this case would occur. If nothing else we should properly handle this contract to be defensive. Some feature we implement a year from now could depend on this and because Windows handles the contract properly it works, but then we have a bug in Mac we have to find and fix.

CartBlanche: I'm not looking to cut corners Eric, but trying to establish real world situations. 
I agree that the contract should be upheld in case of future dependency.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
CartBlanche: Ok

Xamarin.PropertyEditing.Mac/Controls/StringEditorControl.cs
CartBlanche: It was the only editor hooked up this way as I'd overlooked a required refactor. Should be fixed in my next push.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
CartBlanche: hmm I may need to talk to you about doing the late creation of this.

CartBlanche:
@ermau @bleroy This PR has now been rebased off the Mac fixes in master.

Xamarin.PropertyEditing.Mac/Controls/Custom/BasePopOverControl.cs
```diff
@@ -0,0 +1,46 @@
...
+		const int DefaultIconButtonSize = 24;
+
+		public BasePopOverControl (string title, string imageNamed)
+		{
+			if (title == null)
+				throw new ArgumentNullException (nameof (title));
+			if (imageNamed == null)
+				throw new ArgumentNullException (nameof (imageNamed));
+
+			Frame = new CGRect (CGPoint.Empty, new CGSize (320, 180));
```
bleroy: Lots of hard-coded constants in there. Are these considered defaults that derived controls will override?

Xamarin.PropertyEditing.Mac/Controls/Custom/BasePopOverViewModelControl.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class BasePopOverViewModelControl : BasePopOverControl
+	{
+		internal IEnumerable<object> Targets => this.viewModel.Editors.Select (ed => ed.Target);
+
+		internal IPropertyInfo Property => this.viewModel.Property;
+
+		private PropertyViewModel viewModel;
```
bleroy: You don't need the field here, you can just as well do `internal PropertyViewModel ViewModel {get;}`.

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
```diff
@@ -0,0 +1,59 @@
...
+			set {
+				if (this.command != null)
+					this.command.CanExecuteChanged -= CanExecuteChanged;
+
+				this.command = value;
+				this.command.CanExecuteChanged += CanExecuteChanged;
+			}
+		}
+
+		public CommandMenuItem (string title) : base (title)
```
bleroy: Can you move constructors up?

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
```diff
@@ -0,0 +1,59 @@
...
+		{
+			HookUpCommandEvents ();
+		}
+
+		public CommandMenuItem (string title, ICommand command) : this (title)
+		{
+			this.Command = command;
+
+			HookUpCommandEvents ();
```
bleroy: You already did that by calling `this(title)`

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
```diff
@@ -0,0 +1,59 @@
...
+			if (this.command != null)
+				return this.command.CanExecute (null);
+
+			return false;
+		}
+
+		private void CanExecuteChanged (object sender, EventArgs e)
+		{
+			var cmd = sender as ICommand;
+			if(cmd != null)
```
bleroy: nit: you can do `if (sender is ICommand cmd) {`

Xamarin.PropertyEditing.Mac/Controls/Custom/BasePopOverControl.cs
CartBlanche: Yes, the Frame will change dependent on each derived View. I could have left the Frame as 0s, but want devs to at least see something once they fire up their derived View.

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
CartBlanche: Nicely spotted. I must have thought I was still calling base there.

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
CartBlanche: Sorry, I'm not sure what you mean here?

Xamarin.PropertyEditing.Mac/Controls/Custom/BasePopOverViewModelControl.cs
CartBlanche: Yeah it is a little overkill. I was thinking it would be needed down the line, but no point pre-empting that.

CartBlanche: Actually thinking about it, these BasePopOverControl changes should go in a separate PR as they don't directly affect PropertyReset, but the PRs that are coming based off PropertyReset. I'll create a new stand alone PR for these changes.

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
CartBlanche: I've not been keeping up with latest language improvements. That is nice!!

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
bleroy: I mean above the Command property.

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
bleroy: It is! :D

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
CartBlanche: Oh, I always tend to have variables/properties etc above constructors, methods and the like. Not always but mostly. Is having the constructors at the top part of our coding standards? If so I've missed that the last 5 years, or is it a MS specific thing?

bleroy: Not coding standards as far as I know, but usage in the rest of the codebase, certainly.

CartBlanche: @bleroy ^ in case you missed my reply.

CartBlanche:
@ermau This has now been rebased off master.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -104,4 +107,13 @@ protected virtual void OnPropertyChanged (object sender, System.ComponentModel.P
 
 		protected abstract void UpdateAccessibilityValues ();
 	}
+
+	internal abstract class PropertyEditorControl<T> : PropertyEditorControl
```
ermau: It'd be better to do something like:
```csharp
internal abstract class PropertyEditorControl<TViewModel>
    : PropertyEditorControl
    where TViewModel : PropertyViewModel
```

Some property types have viewmodels that subclass from the standard `PropertyViewModel<T>`, if we just let them declare the VM type they expect, they won't need to do this work _again_.

CartBlanche: Good point.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -0,0 +1,102 @@
...
+				case ValueSource.Style:
+				case ValueSource.Resource:
+					ToolTip = null;
+					break;
+			}
+		}
+
+		private void OnPropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
+		{
+			if (e.PropertyName == "ValueSource") {
```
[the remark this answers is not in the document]
CartBlanche: Hmm thought I'd already done that. Obviously not. I'll fix that now.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -0,0 +1,102 @@
...
+					break;
+
+				case ValueSource.Inherited:
+					AlternateImage = NSImage.ImageNamed ("property-button-inherited-mac-active-10");
+					Image = NSImage.ImageNamed ("property-button-inherited-mac-10");
+					ToolTip = Properties.Resources.Inherited;
+					break;
+				case ValueSource.DefaultStyle:
+				case ValueSource.Style:
+				case ValueSource.Resource:
```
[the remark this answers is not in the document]
CartBlanche: Sure.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
CartBlanche: @ermau What I was going to ping you about was.
1. Shouldn't the TViewModel above be of type PropertyViewModel<T>?
2. Are you suggesting this new class will be used across both Mac and Windows?
3. When you say "let them declare the VM type they expect" are you referring to when they register the mapping between VM and Type?

I think I'm misunderstanding exactly what you mean or possibly over complicating it in my head.

ermau: 1. It is: `where TViewModel : PropertyViewModel`
2. No.
3. No. The individual editor implementations can declare the VM they expect to be dealing with when they subclass `PropertyEditorControl<TViewModel>`. Ex: `class PredefinedValuesEditor<T> : PropertyEditorControl<PredefinedValuesViewModel<T>>`

ermau: Still no, `ObjectPropertyViewModel` does not use the strongly typed version for example.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -411,6 +411,11 @@ public ICommand NavigateToValueSourceCommand
 
 		public virtual bool SupportsValueSourceNavigation => false;
 
+		public virtual ValueSource ValueSource
```
ermau: This should have been left abstract, its value must be valid in all cases, an implementing subclass can not leave this as default.

Xamarin.PropertyEditing/ViewModels/ObjectPropertyViewModel.cs
```diff
@@ -107,11 +107,10 @@ private set
 			}
 		}
 
-		public ValueSource ValueSource
+		public new ValueSource ValueSource
```
ermau: Instead of using `new` here and making this now an ambiguous second version of a property on the base class a better solution would be to move the `private set` outside the property into a separate method. Since it's private here we don't actually need it as part of the property; the mental cost of having it a separate (ex.) `SetValueSource` method is lower than that of tracking two separate versions of the same property.

`new` should almost never be used on member declarations, its usage should be considered very carefully.

CartBlanche: The original code from master didn't have the `new` there, so was added to remove the compiler warning without refactoring the `private set` to another method. I'll refactor it out now we're making the base property abstract.

PR #144, opened 2018-01-30
Basic Windows Resources

ermau:
This PR:
 - Flushes out the resources API
 - Adds basic brush preview templates (useful for the brush editor)
 - Adds documentation and clarifies some SetValue behavior
 - Adds the basis for a resource selector window

While we don't yet have a design for a resource selector window, the basics of one largely won't change from the internals perspective. As such this adds a basic one based on the type selector window that we can adjust once we have designs. The complete experience of the standin helps to test and make sure the rest of the internals are working as expected.

### DIFF
--- a/Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
+++ b/Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
@@ -26,7 +26,7 @@ namespace Xamarin.PropertyEditing.Tests
 			return Task.FromResult<IReadOnlyList<ResourceSource>> (new[] { SystemResourcesSource });
 		}
 
-		private static readonly ResourceSource SystemResourcesSource = new ResourceSource ("System Resources", isLocal: false);
+		internal static readonly ResourceSource SystemResourcesSource = new ResourceSource ("System Resources", isLocal: false);
 
 		private static readonly IReadOnlyList<Resource> SystemResources = new Resource[] {
 			new Resource<CommonSolidBrush> (SystemResourcesSource, "ControlTextBrush", new CommonSolidBrush (0, 0, 0)),
--- a/Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
+++ b/Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
@@ -31,6 +31,7 @@ namespace Xamarin.PropertyEditing.Tests
 		private static readonly IReadOnlyList<Resource> SystemResources = new Resource[] {
 			new Resource<CommonSolidBrush> (SystemResourcesSource, "ControlTextBrush", new CommonSolidBrush (0, 0, 0)),
 			new Resource<CommonSolidBrush> (SystemResourcesSource, "HighlightBrush", new CommonSolidBrush (51, 153, 255)),
+			new Resource<CommonSolidBrush> (SystemResourcesSource, "TransparentBrush", new CommonSolidBrush (0, 0, 0, 0)),
 			new Resource<CommonColor> (SystemResourcesSource, "ControlTextColor", new CommonColor (0, 0, 0)),
 			new Resource<CommonColor> (SystemResourcesSource, "HighlightColor", new CommonColor (51, 153, 255))
 		};
--- a/Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
+++ b/Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
@@ -1,0 +1,38 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xamarin.PropertyEditing.Drawing;
+
+namespace Xamarin.PropertyEditing.Tests
+{
+	public class MockResourceProvider
+		: IResourceProvider
+	{
+		public Task<Resource> CreateResourceAsync<T> (ResourceSource source, string name, T value)
+		{
+			throw new NotImplementedException ();
+		}
+
+		public Task<IReadOnlyList<Resource>> GetResourcesAsync (object target, IPropertyInfo property, CancellationToken cancelToken)
+		{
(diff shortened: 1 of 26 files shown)


Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+		{
+			return Task.FromResult<IReadOnlyList<Resource>> (SystemResources.Where (r => property.Type.IsAssignableFrom (r.GetType().GetGenericArguments()[0])).ToList());
+		}
+
+		public Task<IReadOnlyList<ResourceSource>> GetResourceSourcesAsync (object target, IPropertyInfo property)
+		{
+			return Task.FromResult<IReadOnlyList<ResourceSource>> (new[] { SystemResourcesSource });
+		}
+
+		internal static readonly ResourceSource SystemResourcesSource = new ResourceSource ("System Resources", isLocal: false);
```
bleroy: We shouldn't hard-code the actual resources. The client code, whether it's test case or the standalone apps, should set those.

Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
```diff
@@ -522,6 +525,30 @@
 		</Setter>
 	</Style>
 
+	<local:CommonBrushToBrushConverter x:Key="BrushConverter" />
+	<local:CommonColorToColorConverter x:Key="ColorConverter" />
+
+	<DataTemplate x:Key="SimpleResourcePreviewItem">
+		<DataTemplate.Resources>
+			<DataTemplate DataType="{x:Type drawing:CommonColor}">
+				<Rectangle>
```
bleroy: This will not show transparency. You should be using `BrushBoxControl`. You should even be able to handle a preview of any brush, in addition to solid colors.

Xamarin.PropertyEditing/Resource.cs
```diff
@@ -0,0 +1,97 @@
+using System;
+
+namespace Xamarin.PropertyEditing
+{
+	public class Resource<T>
```
bleroy: One thing you can find in my branch is covariance on `T`, which is useful to handle setting a `Brush` property with a `Resource<SolidBrush>` resource. In particular, a `GetResourcesAsync (obj, brushPropertyInfo, token)`can give you all resources with a representation type that derives from `Brush`, and the client code can then manipulate those resources as `Resource<Brush>` much more easily.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -181,49 +155,19 @@ private void OnSetValueToResource (Resource resource)
...
-				Source = ValueSource.Default,
+				Source = clearSource,
 				Value = default(TValue)
 			});
 		}
 
 		private bool CanClearValue ()
 		{
-			return (Property.ValueSources.HasFlag (ValueSources.Local) && Property.ValueSources.HasFlag (ValueSources.Default) && ValueSource == ValueSource.Local);
+			return (ValueSource != ValueSource.Default && ValueSource != ValueSource.Unknown);
```
bleroy: Aha! :)

Xamarin.PropertyEditing/ViewModels/ResourceSelectorViewModel.cs
```diff
@@ -0,0 +1,144 @@
...
+		{
+			if (String.IsNullOrWhiteSpace (FilterText) && ShowSystemResources) {
+				this.resourcesView.Options.Filter = null;
+				return;
+			}
+
+			this.resourcesView.Options.Filter = ResourceFilter;
+		}
+
+		private bool ResourceFilter (object item)
```
bleroy: Why isn't that taking `Resource` since you're going to immediately cast it anyway?

ermau: Because that would just move the cast somewhere else and create a second method in the process.

Xamarin.PropertyEditing.Tests/MockControls/MockResourceProvider.cs
ermau: Test cases should largely be using inline `Mock<>`, not this. The standalone apps I'm not really worried about being that flexible, we can adapt it later if need be.

Xamarin.PropertyEditing/ViewModels/ResourceSelectorViewModel.cs
bleroy: It would move it arguably where it belongs, with a safer result: we're trading compile-time safety for runtime exceptions. I'm not following why this would create a second method.

ermau: > we're trading compile-time safety for runtime exceptions.

No we're not, we'll have to cast it regardless, `SimpleCollectionView` is not generic remember?

> I'm not following why this would create a second method.

3 lines up.

bleroy: Well, it's private anyways.

Xamarin.PropertyEditing/Resource.cs
ermau: It should be unnecessary.

> which is useful to handle setting a Brush property with a Resource<SolidBrush> resource.

The type of `Resource` actually doesn't matter for setting brush properties, it only cares about the base type, the provider will then recycle it back up into a `Resource<Brush>` when it's retrieved.

>  In particular, a GetResourcesAsync (obj, brushPropertyInfo, token)can give you all resources with a representation type that derives from Brush, and the client code can then manipulate those resources as Resource<Brush> much more easily.

```csharp
Resource<SolidBrush> resource;
// Why is
IResource<Brush> basic = resource;
DoStuff (basic);
// Better than
DoStuff (resource.Value);
```

We'd be adding an interface to support something we generally don't need in the first place, I'm unconvinced.

bleroy: It's better wherever the client code is manipulating any brush, and not a concrete brush, solid or otherwise. I added it in the first place because while experimenting with using the API, it did make things easier imo. But that's fine, we can very easily add it later if we find it an improvement. That you haven't hit any issue with that PR's code shows that it's not really needed now. It's a suggestion.

PR #146, opened 2018-01-31
Improve accessibiltiy of the thickness properties

### LINKED ISSUE #124: ThicknessEditorControl does not include accessible labels for properties
The thickness editor control appears to use only paths as labels.  They need to at least be marked up with text names for accessibility purposes

lewing:
The thickness properties had no accessible name, this
adds AutomationProperties.Name to the paths to improve
accessibility.  Fixes #124


ermau:
This just seems to add names to the icons? The actual entry fields already have them, were they not working?

PR #152, opened 2018-02-02
[build] Add Provisionator so we can build for references to ValidateM…

alanmcgovern:
…enuItem on Mac

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+/bot-provisioning/_provisionator/
+/bot-provisioning/provisionator.msbuild.props
 packages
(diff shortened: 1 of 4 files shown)


alanmcgovern:
Perfect! I just stripped out the `xcode.csx` one because that's specific to the ios designer test suite :)

CartBlanche:
I was hoping to test this on the PropertyReset branch, first and it didn't build there :/

alanmcgovern:
if you need a newer/older build of xammac then we can update the version in the csx. The actual provisioning part is good!

PR #179, opened 2018-02-12
[Core] Handle numerics more generically

ermau:
While the editors themselves still handle things at the wider level, at
least now the core VMs use the real type via normal conversion. This
should allow editor implementers to use the type they expect without
issue.


Xamarin.PropertyEditing/Numeric.cs
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		private static Func<T, T> IncrementValue, DecrementValue;
+
+		private static void SetupMath ()
+		{
+			if (IncrementValue != null)
+				return;
+
+			Task<Func<T, T>> createIncrement = Task.Run (() => {
```
bleroy: Hehe, you knew I'd comment on that. In addition to the pattern matching solution that's been suggested elsewhere, I don't see why you shouldn't also be able to setup the increment and decrement delegates early on to point to Lambdas written in plain C# that are specialized for each type, but can be cast to your generic signature.
Then again, as far as IL gen goes, this is not the worst ;)

ermau: Frankly it was because I thought I could get away with fewer branches than this originally and that'd make it more concise than a full method for every type. Mainly just stayed on the train of knowing the approach would work one way or another.

Issue #136, opened 2018-01-23
Windows Material color selector


ermau:
- [x] Design
  - I think the most logical thing here is to place it as another "color type" tab:
![image](https://user-images.githubusercontent.com/156582/35292688-4c76d570-003f-11e8-8f76-d5f9dbcc9b71.png)

 - [ ] Localized
 - Behavior:
   - [ ] Keyboard control
   - [ ] Accessibility
   - [ ] Selection
 - Theming:
   - [ ] Light themed
   - [ ] Dark themed
   - [ ] High contrast

vancura:
## Inactive state

Nothing is selected.

<img width="408" alt="win android material color inline palette 2x" src="https://user-images.githubusercontent.com/4982/35618690-f9cbf6b4-067b-11e8-97fa-fdf5a60a88a3.png">

## Active Material Editor

<img width="408" alt="win android material color inline palette 2x" src="https://user-images.githubusercontent.com/4982/35618725-160918e8-067c-11e8-881b-abfb8075388e.png">

## Dark inactive state

<img width="408" alt="win android material color inline palette _dark 2x" src="https://user-images.githubusercontent.com/4982/35618774-3ac065ce-067c-11e8-9f25-d01fd60b4db8.png">

## Dark active state

<img width="408" alt="win android material color inline palette _dark 2x" src="https://user-images.githubusercontent.com/4982/35618811-5106ccba-067c-11e8-9b35-a85ac809373f.png">

Thoughts?

PR #180, opened 2018-02-13
Material design color picker with Windows implementation

bleroy:
materialdesigncolorpicker

Known issues:

* [ ] Styling is not 100% the planned design from #136 (width of the popup doesn't grow when material design tab is selected, for instance)
* [ ] Tab icons are not final
* [x] There are binding errors popping up in the output window when running the standalone app
* [x] It takes two clicks for changes to fully register (first time, selected style may not apply, or selection may not be acknowledged; I believe the problem is in `ChoiceControl` when `SelectedItem` is bound rather than `SelectedValue`)
* [x] Black and White colors are not in the design. They should be, will follow-up on what the design should be for those.


Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
```diff
@@ -1113,6 +1190,150 @@
 		</Setter>
 	</Style>
 
+	<Style x:Key="BrushChoiceControlItem" BasedOn="{StaticResource {x:Type RadioButton}}" TargetType="RadioButton">
```
ermau: Can we name this more specifically? It's meant for the material picker right? It sounds like its for selecting the brush type.

ermau:
>  It takes two clicks for changes to fully register (first time, selected style may not apply, or selection may not be acknowledged

This must be fixed.

Xamarin.PropertyEditing.Windows/BrushTabbedEditorControl.cs
```diff
@@ -83,14 +115,43 @@ private void StorePreviousBrush ()
...
+				ShowSelectedTab ();
 				break;
 			case CommonSolidBrush _:
 				this.brushChoice.SelectedValue = solid;
+				ShowSelectedTab ();
+				break;
+			}
+		}
+
+		private void ShowSelectedTab()
```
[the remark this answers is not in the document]
bleroy: That will be part of the changes we talked about on the view model, I didn't attempt to fix that in this PR.

Xamarin.PropertyEditing/Drawing/CommonColor.cs
```diff
@@ -43,6 +44,10 @@ public CommonColor (byte r, byte g, byte b, byte a = 255)
 		/// Blue component
 		/// </summary>
 		public byte B { get; }
+		/// <summary>
+		/// An optional label for the color, that does not affect equality or anything else.
```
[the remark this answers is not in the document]
bleroy: Good point.

Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
bleroy: Sure.

bleroy:
Definitely. The rest as well.

Xamarin.PropertyEditing/ViewModels/MaterialDesignColorViewModel.cs
```diff
@@ -0,0 +1,251 @@
...
+				.OrderBy (palette => palette.Colors.Min (
+					paletteColor => CommonColor.SquaredDistance (paletteColor, Color)))
+				.First ();
+
+		public CommonColor? NormalColor
+		{
+			get {
+				if (this.normalColor.HasValue) return this.normalColor.Value;
+				MaterialColorScale scale = FindPalette (ColorName, false);
+				foreach (CommonColor normalColor in scale.Colors) {
```
lewing: It looks like FindPalette can return null so this needs to handle that. Actually there are a lot of places where FindPalette results are accessed without a check?

bleroy: Yes, you're right.

Xamarin.PropertyEditing/Drawing/CommonColor.cs
```diff
@@ -8,12 +8,13 @@ namespace Xamarin.PropertyEditing.Drawing
 	[Serializable]
 	public struct CommonColor : IEquatable<CommonColor>
 	{
-		public CommonColor (byte r, byte g, byte b, byte a = 255)
+		public CommonColor (byte r, byte g, byte b, byte a = 255, string name = "")
```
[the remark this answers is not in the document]
bleroy: No particular reason, null works just as well. Changing it.

Xamarin.PropertyEditing/ViewModels/MaterialDesignColorViewModel.cs
lewing: 👍

PR #186, opened 2018-02-14
[Core/Win] Combinable editor

ermau:


Xamarin.PropertyEditing.Tests/CombinablePredefinedViewModelTests.cs
```diff
@@ -0,0 +1,383 @@
...
+			Assume.That (flag4Choice.Value, Is.EqualTo ((int)FlagsTestEnum.Flag4));
+			var flag5Choice = vm.Choices.Single (v => v.Name == nameof(FlagsTestEnum.Flag5));
+			Assume.That (flag5Choice.Value, Is.EqualTo ((int)FlagsTestEnum.Flag5));
+
+			Assert.That (vm.Value, Is.EqualTo (default(int)));
+			Assert.That (vm.MultipleValues, Is.True, "MultipleValues is not marked true for disagreeing values");
+			Assert.That (flag2Choice.IsFlagged.HasValue && flag2Choice.IsFlagged.Value, Is.True);
+			Assert.That (flag3Choice.IsFlagged.HasValue, Is.False, "Disagreed upon choice is not indeterminate");
+			Assert.That (flag4Choice.IsFlagged.HasValue, Is.False, "Disagreed upon choice is not indeterminate");
+			Assert.That (flag5Choice.IsFlagged.HasValue, Is.False, "Disagreed upon duplicate choice is not indeterminate");
```
bleroy: When you have the time, I'd add to the comment, as this is confusing: since 5 has the same value as 4, you'd expect the behavior to be the same, except that one is explicitly included in the actual value during setup, and the other is sort of an accident. If I got that correctly...

Xamarin.PropertyEditing.Tests/CombinablePredefinedViewModelTests.cs
```diff
@@ -0,0 +1,383 @@
...
+			Assume.That (vm.Value, Is.EqualTo (default(int)));
+		
+			vm.Choices[1].IsFlagged = true;
+			Assert.That (setValue, Is.Not.Null, "Did not call setvalue");
+			CollectionAssert.AreEquivalent (new[] { (int)FlagsTestEnum.Flag1 }, setValue.Value);
+			Assert.That (vm.Value, Is.EqualTo ((int) value));
+		}
+
+		[Test]
+		[Ignore ("There's either a bug in the test or in Moq, but the callbacks seem to be causing NREs even empty")]
```
bleroy: ouch

PR #214, opened 2018-02-21
Bleroy md fit and finish

bleroy:
Small fixes in visual appearance of the material design color picker, plus a small NRE fix


Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
```diff
@@ -1167,9 +1167,17 @@
 		</Setter>
 		<Style.Triggers>
 			<Trigger Property="IsChecked" Value="True">
-				<Setter Property="BorderBrush" Value="{DynamicResource ToggleItemSelectedBorderBrush}" />
-				<Setter Property="BorderThickness" Value="1"/>
-				<Setter Property="Padding" Value="1"/>
+				<Setter Property="Template">
```
ermau: Replacing the template is not the way to go here, this should have the exact same effect but at worse performance. I'd recommend having the original template such that the padding and border thickness do not change, only the color.

bleroy: Fair point, but that won't work consistently for the padding. Instead, I moved the triggers inside the template, and targeted the setters to the border.

PR #224, opened 2018-02-22
Check that ValueSources contains ValueSource.Resource

CartBlanche:


Xamarin.PropertyEditing/BrushPropertyInfo.cs
```diff
@@ -7,7 +7,7 @@ namespace Xamarin.PropertyEditing
 	public class BrushPropertyInfo : IPropertyInfo, IColorSpaced
 	{
 		public BrushPropertyInfo (string name, string category, bool canWrite,
-			IReadOnlyList<string> colorSpaces = null, ValueSources valueSources = ValueSources.Local | ValueSources.Default,
+			IReadOnlyList<string> colorSpaces = null, ValueSources valueSources = ValueSources.Resource | ValueSources.Default,
```
CartBlanche: @bleroy Is this change correct for the Resources functionality. Do we need these on any other *PropertyInfo classes like for string resources, for example.

PR #225, opened 2018-02-24
Added support for theming proppy via VS color theme

BretJohnson:
This code is packaged with proppy source, so we can keep the
color list up to date here, but it's intended to be compiled with
the client app (which is straightforward - the VS Designer code just
points to this source file in the proppy git submodule).

Everything seems to work well here, except that there are
some brush names that I couldn't find matches for in the XET
source. Those brushes are commented out in AddBrushes. @ermau -
for those brushes, the ones commented out, can you let me know
where you got those brush names / colors from. I'll then update
the mapping to handle them too.

### DIFF
--- a/VisualStudioIntegration/VisualStudioTheme.cs
+++ b/VisualStudioIntegration/VisualStudioTheme.cs
@@ -43,26 +43,26 @@ namespace Xamarin.PropertyEditing.VisualStudioIntegration
 	/// </example>
 	public class VisualStudioTheme
 	{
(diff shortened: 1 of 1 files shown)


ermau:
Until we have a clear plan of this piece's place inside the proppy repository it doesn't belong here, but in the individual integration. A loose file that can has no way of compiling it has no place here. Furthermore this is a manual list to maintain when there are clear automated ways.

PR #237, opened 2018-03-02
Fixes to Support Nullables On Mac

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,144 @@
...
+			}
+
+			base.OnViewModelChanged (oldModel);
+		}
+
+		protected override void UpdateValue ()
+		{
+			foreach (var item in combinableList) {
+				if (ViewModel.Choices.Count () > 0) {
+					var flag = ViewModel.Choices.Where (x => x.Name == item.Title).FirstOrDefault ();
```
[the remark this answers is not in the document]
CartBlanche: @bleroy meant to tidy that up after writing it but got distracted with other fixes and forgot. Nicely spotted.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -17,6 +17,11 @@ internal class PropertyTableDelegate
 		public PropertyTableDelegate (PropertyTableDataSource datasource)
 		{
 			this.dataSource = datasource;
+
+			// Add the Mac specific mappings
```
ermau: These shouldn't be added back in.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -462,7 +462,7 @@ private PropertyViewModel GetViewModelCore (IPropertyInfo property)
 
 		private Task busyTask;
 
-		private static readonly Dictionary<Type,Func<TargetPlatform,IPropertyInfo,IEnumerable<IObjectEditor>,PropertyViewModel>> ViewModelMap = new Dictionary<Type, Func<TargetPlatform, IPropertyInfo, IEnumerable<IObjectEditor>, PropertyViewModel>> {
+		internal static readonly Dictionary<Type,Func<TargetPlatform,IPropertyInfo,IEnumerable<IObjectEditor>,PropertyViewModel>> ViewModelMap = new Dictionary<Type, Func<TargetPlatform, IPropertyInfo, IEnumerable<IObjectEditor>, PropertyViewModel>> {
```
ermau: We don't need to add the mac specific types back in so this can stay private.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -109,23 +113,30 @@ PropertyEditorControl GetEditor (EditorViewModel vm, NSOutlineView outlineView)
...
-			if (controlType.IsGenericTypeDefinition) {
-				if (genericArgs == null)
-					genericArgs = propertyType.GetGenericArguments ();
+				return SetUpEditor (controlType, vm, outlineView);
+			} else
+				return null;
+		}
 
-				controlType = controlType.MakeGenericType (genericArgs);
+		private Type[] GetCorrectGenericArguments (Type[] types)
```
ermau: This shouldn't be needed, as it stands the properties VM will treat all numerics as nullables, so the controls should just deal with them as such. That's why `Numeric` can handle `T` as a `Nullable<>`

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,144 @@
...
+				dataPopulated = true;
+			}
+
+			base.OnViewModelChanged (oldModel);
+		}
+
+		protected override void UpdateValue ()
+		{
+			foreach (var item in combinableList) {
+				if (ViewModel.Choices.Count () > 0) {
```
ermau: `Any()` is faster than `Count() > 0`. Beyond that, `Choices` is an `IReadOnlyList`, we don't need to go to LINQ to see if it's empty.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,144 @@
...
+				SetErrors (errors);
+			} else {
+				SetErrors (null);
+				SetEnabled ();
+			}
+		}
+
+		protected override void OnViewModelChanged (PropertyViewModel oldModel)
+		{
+			if (!dataPopulated) {
```
ermau: So when the view model changes, but it's already been loaded once, how does it load the new view model's list of choices?

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,144 @@
...
+
+				combinableList.Clear ();
+
+				const float controlHeight = 22;
+
+
+				// Set our new RowHeight
+				RowHeight =  (ViewModel.Choices.Count () * controlHeight);
```
ermau: No need for LINQ Count.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
CartBlanche: Agreed. I added them back in so I could test the AppKit, but if we aren't supporting that anymore.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
CartBlanche: Ok

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
CartBlanche: I hadn't see the Count property on IReadOnlyList. Much nicer.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
CartBlanche: @ermau This is needed at the point where we are dynamically creating the control, before we insert it into the TableView. 
`controlType.MakeGenericType (genericArgs);` will fail/crash if we try to create the type with a Nullable<>. If you know of a way around that, then I'm all ears.

ermau: Because you're constraining it to a struct, `Nullable` isn't a struct.

CartBlanche: @ermau I think I misunderstand what you mean, because according to the docs `Nullable<T>` is a struct - https://msdn.microsoft.com/en-us/library/b3h38hb0(v=vs.110).aspx

ermau: No, I'm just misremembering an error I ran into related to this before. I need the crash info if I'm to help instead of just guessing.

CartBlanche: I think this is all the info you might need - https://gist.github.com/CartBlanche/34a918b06d77fabdfec87112088ad22b
Let me know if I'm missing any info you need.

CartBlanche:
@ermau I'm rebased off master and everything appears to be hunky dory again.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -115,17 +114,16 @@ PropertyEditorControl GetEditor (EditorViewModel vm, NSOutlineView outlineView)
...
-			if (controlType == null)
-				return null;
-
-			if (controlType.IsGenericTypeDefinition) {
-				if (genericArgs == null)
-					genericArgs = propertyType.GetGenericArguments ();
-
-				controlType = controlType.MakeGenericType (genericArgs);
-			}
+			if (controlType != null) {
```
ermau: Needless changes

CartBlanche: Reverted.

PR #256, opened 2018-03-15
[Win] Add basic property description support

ermau:


Xamarin.PropertyEditing.Tests/MockPropertyInfo/MockPropertyInfo.cs
```diff
@@ -26,6 +26,7 @@ public MockPropertyInfo (string name, string category = "", bool canWrite = true
 		}
 
 		public string Name { get; }
+		public string Description => null;
```
bleroy: Let's have an optional constructor parameter for it and actually have it do something in the mock property info, that's useful to have in the standalones.

PR #258, opened 2018-03-15
Validation/Coerce

ermau:
This PR:
 - Renames the previous validation to coercion as that's what it is
 - Adds validation support at a base level and special none handling for flags
 - Fixes previous failing behavior for flags multi-select

### DIFF
--- a/Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
+++ b/Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
@@ -177,6 +177,19 @@ namespace Xamarin.PropertyEditing.Tests
 					softType.GetProperty ("Value").SetValue (softValue, v);
 					softType.GetProperty ("Source").SetValue (softValue, value.Source);
 				}
+
+				if (typeof(T).Name == "IReadOnlyList`1") {
+					var list = (IReadOnlyList<int>) value.Value;
+					int iv = 0;
+					foreach (int flag in list) {
+						iv |= flag;
+					}
+
+					softValue = new ValueInfo<int> {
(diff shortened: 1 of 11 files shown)


Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
```diff
@@ -177,6 +177,19 @@ public Task<IReadOnlyList<ITypeInfo>> GetAssignableTypesAsync (IPropertyInfo pro
 					softType.GetProperty ("Value").SetValue (softValue, v);
 					softType.GetProperty ("Source").SetValue (softValue, value.Source);
 				}
+
+				if (typeof(T).Name == "IReadOnlyList`1") {
+					var list = (IReadOnlyList<int>) value.Value;
```
bleroy: Looks fine to start with just a list of ints, but I'd still put some safeguards in there and at least throw not implemented if that cast fails.

Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
```diff
@@ -209,9 +224,34 @@ public Task<IReadOnlyList<ITypeInfo>> GetAssignableTypesAsync (IPropertyInfo pro
 					};
 				} else if (value == null || value is T) {
 					return new ValueInfo<T> {
-						Value = (T)value,
+						Value = (T) value,
 						Source = ValueSource.Local
 					};
+				} else if (tType.Name == "IReadOnlyList`1") {
+					// start with just supporting ints for now
```
bleroy: Same here. Maybe an assertion?

PR #261, opened 2018-03-21
[Mac]Initial implementation for Combinable Property Editor

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,137 @@
...
+				item.AccessibilityEnabled = item.Enabled;
+				item.AccessibilityTitle = string.Format (LocalizationResources.AccessibilityCombobox, ViewModel.Property.Name);
+			}
+		}
+
+		void SelectionChanged (object sender, EventArgs e)
+		{
+			var button = sender as NSButton;
+
+			if (button != null) {
```
bleroy: nit: `if (sender is NSButton button) {`

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,137 @@
...
+				var choice = ViewModel.Choices.First (x => x.Name == button.Title);
+				switch (button.State) {
+					case NSCellStateValue.Off:
+						choice.IsFlagged = false;
+						break;
+					case NSCellStateValue.On:
+						choice.IsFlagged = true;
+						break;
+				}
```
bleroy: Q: in mixed state, which is not explicitly handled here, are we even getting here, and what should we be doing? Maybe add a comment.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
CartBlanche: hehe I didn't review it after pulling out the code from the other PR. I'll add that in shortly.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
CartBlanche: My understanding is that it can only ever be in a mixed state by either reset, or first time we load. So If a selection is made it can only be on or off. Am I making sense?

bleroy: Totally, that was kind of my assumption, but I wasn't sure. Thanks.

Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
```diff
@@ -52,9 +52,9 @@ public enum Enumeration
 		[Flags]
 		public enum Flags
 		{
-			FlagOne,
-			FlagTwo,
-			FlagThree
+			FlagOne = 1,
```
CartBlanche: If we don't add values to the Flagged Enum, it treats the first Enum as 0, which then messes up all toggling logic and therefore doesn't work correctly.

ermau: No, it's supposed to be able to handle a 0, leave this as is.

CartBlanche:
@ermau The behaviour now is more consistent across Mac and Windows for both cases of if the Flags have values and when they don't.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,142 @@
...
+			firstKeyView = combinableList.First ();
+			lastKeyView = combinableList.Last ();
+
+			base.OnViewModelChanged (oldModel);
+		}
+
+		protected override void UpdateValue ()
+		{
+			foreach (var item in combinableList) {
+				if (ViewModel.Choices.Count > 0) {
```
ermau: How would it be possible for there to be no choices in the VM, but for `combinableList` to have content?

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,142 @@
...
+					var flag = ViewModel.Choices.FirstOrDefault (x => x.Name == item.Title);
```
ermau: Why not just have `combinableList` be a `Dictionary` mapping them so you don't need to do a string compare lookup everytime?

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
```diff
@@ -0,0 +1,142 @@
...
+					if (flag.IsFlagged.HasValue) {
+						item.State = flag.IsFlagged.Value ? NSCellStateValue.On : NSCellStateValue.Off;
+						item.AllowsMixedState = false;
```
ermau: You're disabling mixed state here, but not re-enabling it below. A value that differs when a second object selected will not show as mixed like it should.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
ermau: > My understanding is that it can only ever be in a mixed state by either reset, or first time we load. 

Almost, another case is if a second object is selected that an individual flag is set for one but not the other.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
CartBlanche: YAR, not possible. I think I was being overly cautious.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
CartBlanche: I thought that once the value as set (either ticked or unticked), it could never be null again, hence why it could never be mixed again?

CartBlanche: The initial state of the NSButton is always mixed, when the button is first created. So by the time UpdateValue is called if it HasValue is false it will already be in a Mixed State. I could set the State to Mixed on instantiating the NSButton and avoid this else.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
CartBlanche: Sure.

Xamarin.PropertyEditing.Mac/Controls/CombinablePropertyEditor.cs
ermau: You're missing the point, the state can become mixed _after_ first setup:
> A value that differs when a second object selected will not show as mixed like it should.

> ... another case is if a second object is selected that an individual flag is set for one but not the other.

If you select the first mock, select flag 1, unselect it, select the second mock and select flag 2 and then select the first mock they should both show mixed, but they both show checked.

PR #267, opened 2018-03-26
Slight refactor to avoid duplicate Identifier code.

CartBlanche:


Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -189,6 +191,24 @@ private PropertyEditorControl SetUpEditor (Type controlType, EditorViewModel pro
 			return view;
 		}
 
+		private void GetVMGroupCellItendifiterFromFacade (NSObject item, out PropertyViewModel vm, out IGroupingList<string, EditorViewModel> group, out string cellIdentifier)
+		{
+			var facade = (NSObjectFacade)item;
+			vm = facade.Target as PropertyViewModel;
+			group = facade.Target as IGroupingList<string, EditorViewModel>;
+			cellIdentifier = (group == null) ? vm.GetType ().Name : group.Key;
```
ermau: Really we don't need to identify the group cells individually, they're all the same time, just a text field basically. Since that should update when we apply the new datacontext to it in getting it, it should update.

PR #270, opened 2018-03-26
Remove ILGenerator based code

ermau:
Pre-requisite to moving to .NET standard.

### DIFF
--- a/Xamarin.PropertyEditing/Numeric.cs
+++ b/Xamarin.PropertyEditing/Numeric.cs
@@ -1,6 +1,6 @@
 using System;
 using System.ComponentModel;
-using System.Reflection.Emit;
+using System.Linq.Expressions;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing/Numeric.cs
```diff
@@ -40,110 +49,40 @@ private static void SetupMath ()
...
-				else {
-					Delegate a = add.CreateDelegate (typeof(Func<,>).MakeGenericType (new[] { t, t }));
-					return (v) => {
-						if (v == null)
-							return (T)Activator.CreateInstance (t);
-						else {
-							return (T)a.DynamicInvoke (NullConverter.ConvertFrom (v));
-						}
-					};
+				Expression add;
```
bleroy: nit: it looks a bit weird not to use a ternary operator here (and for the sub expression below).

PR #271, opened 2018-03-29
Accessibility fixes

iainx:
Some improvements to the already outstanding accessibility support.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BaseEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BaseEditorControl.cs
@@ -37,7 +37,6 @@ namespace Xamarin.PropertyEditing.Mac
 			actionButton = new NSButton {
 				Bordered = false,
 				Enabled = false,
-				Hidden = true,
 				ImageScaling = NSImageScale.AxesIndependently,
 				Title = string.Empty,
 				TranslatesAutoresizingMaskIntoConstraints = false,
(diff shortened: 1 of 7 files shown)


Xamarin.PropertyEditing.Mac/Controls/BaseEditorControl.cs
```diff
@@ -19,7 +21,25 @@ public NSButton ActionButton
 			get { return actionButton; }
 		}
 
-		public event EventHandler PropertyButtonClicked;
+		event EventHandler propertyButtonClicked;
+		public event EventHandler PropertyButtonClicked {
+			add {
+				lock (propertyButtonClicked) {
+					propertyButtonClicked += value;
+					propertyButton.Hidden = false;
```
ermau: This doesn't seem related to accessibility changes. We won't be handling whether property buttons are visible or not this way, currently we should assume always visible.

iainx: Fair enough, the reason I did it was buttons that are visible and don't do anything is fairly bad for accessibility, but I can remove it

CartBlanche: @iainx My CustomExpression PR refactors the PropertyButton functionality into its own class. I also agree that Buttons that do nothing, in the case of ReadOnly properties, shouldn't be visible. Showing something that isn't interactive is both bad accessibility and a bad UX.

CartBlanche:
@iainx, @ermau, @lewing I've rebased these changes off of master, which included the PropertyButton refactor. Iain could you please take a look and confirm I've done the right thing and not broken anything??

PR #272, opened 2018-03-29
Change to allow NumericEditor to handle nullables.

CartBlanche:


CartBlanche:
@ermau Please take a look, when you have a mo.

PR #288, opened 2018-04-23
[Mac] Initial CustomExpression Implementation.

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/Custom/BasePopOverViewModelControl.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class BasePopOverViewModelControl : BasePopOverControl
+	{
+		internal IEnumerable<object> Targets => this.ViewModel.Editors.Select (ed => ed.Target);
```
ermau: Doesn't seem that we're using this anywhere?

Xamarin.PropertyEditing.Mac/Controls/Custom/CommandMenuItem.cs
```diff
@@ -24,16 +25,21 @@ public CommandMenuItem (string title) : base (title)
...
 		}
 
 		private void HookUpCommandEvents ()
 		{
 			Activated += (object sender, EventArgs e) => {
-				if (this.command != null)
-					this.command.Execute (null);
+				if (this.command != null) {
+					if (this.command.GetType ().IsGenericType)
+						this.command.Execute (senderView);
```
ermau: Why would we pass UI elements to commands?

CartBlanche: I'm struggling now to remember why I made that change :/ I think it was because I had issues losing the view's handle but I'm handling things differently now and it was a hack to get things working, but isn't needed anymore.

Xamarin.PropertyEditing.Mac/Controls/Custom/BasePopOverViewModelControl.cs
CartBlanche: IIRC this would be used in the RequestResource, but I'll add that in that PR.

PR #289, opened 2018-04-24
Implement ImageBrush editor

bleroy:


Xamarin.PropertyEditing/ViewModels/ImageBrushViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+		public IReadOnlyList<KeyValuePair<CommonImageSource, string>> ImageSources
+		{
+			get => GetImageResourcesAsync ().Result
+				.Where (r => r.Source.IsLocal)
+				.Cast<Resource<CommonImageSource>>()
+				.Select (r => new KeyValuePair<CommonImageSource, string>(r.Value, r.Name))
+				.ToList();
+		}
+
+		public async Task<IReadOnlyList<Resource>> GetImageResourcesAsync()
```
ermau: Why do we need this to be public?

Xamarin.PropertyEditing/ViewModels/ImageBrushViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+		{
+			if (Parent.ResourceProvider == null)
+				return null;
+
+			IReadOnlyList<Resource> resources = await Parent.ResourceProvider.GetResourcesAsync (this, ImageSourceProperty, CancellationToken.None);
```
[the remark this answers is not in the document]
ermau: Why are we querying the outside provider with a property it can't compare to? Why aren't we just querying the actual property and filtering here?

Xamarin.PropertyEditing/ViewModels/ImageBrushViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+			set {
+				if (ImageSource != value) {
+					Parent.Value = (Parent.Value as CommonImageBrush ?? new CommonImageBrush ())
+						.CopyWith (imageSource: value);
+					OnPropertyChanged ();
+				}
+			}
+		}
+
+		public IReadOnlyList<KeyValuePair<CommonImageSource, string>> ImageSources
```
ermau: I realize the assumption here is that the UI will query it once and that's all we need to worry about, but let's treat a property as it should be. We need an async value that we need to do some post-processing on, so let's fire the async request at `ctor` and do the transform task-inline (`ContinueWith` / `ConfigureAwait (false)`.) We should not re-query this end result, it should be cached. In the property we _could_ then block on this final cached task, but it's entirely possible this query time will be significant. There's a type in the library to deal with this exact scenario (`AsyncValue`) to which you can hand a task and bind to the `.Value` of it and have the binding system automatically update when it's finished loading. Additionally it provides a loading indicator that can be bound to.

Xamarin.PropertyEditing/ViewModels/ImageBrushViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+			get => Parent.Value is CommonImageBrush imageBrush ? imageBrush.TileMode : CommonTileMode.None;
+			set {
+				if (TileMode != value) {
+					Parent.Value = (Parent.Value as CommonImageBrush ?? new CommonImageBrush()).CopyWith (tileMode: value);
+					OnPropertyChanged ();
+				}
+			}
+		}
+
+		public IEnumerable<KeyValuePair<CommonTileMode, string>> TileModeValues
```
ermau: We need to consider that the target platforms may not use the same terms exposed in our common types and as such we need to provide some kind of translation here.

Xamarin.PropertyEditing/ViewModels/ImageBrushViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+	internal class ImageBrushViewModel : NotifyingObject
+	{
+		public ImageBrushViewModel (BrushPropertyViewModel parent)
+		{
+			Parent = parent ?? throw new ArgumentNullException (nameof (parent));
+			parent.PropertyChanged += Parent_PropertyChanged;
+		}
+
+		public CommonImageBrush PreviousImageBrush { get; set; }
```
ermau: As a public non-readonly property this needs INPC.

Xamarin.PropertyEditing/ViewModels/BrushPropertyViewModel.cs
```diff
@@ -176,12 +198,16 @@ private void SetBrushType (CommonBrushType type)
...
 				Solid.CommitLastColor ();
 				Solid.CommitHue ();
+				if (type == CommonBrushType.MaterialDesign) {
+					MaterialDesign.SetToClosest ();
+				}
 				break;
-			case CommonBrushType.MaterialDesign:
-				MaterialDesign.SetToClosest ();
+			case CommonBrushType.Image:
+				Value = Image.PreviousImageBrush ?? new CommonImageBrush("");
```
ermau: The `ctor` takes in `null` as a default, but you're passing `""` here, why?

Xamarin.PropertyEditing/ViewModels/BrushPropertyViewModel.cs
```diff
@@ -176,12 +198,16 @@ private void SetBrushType (CommonBrushType type)
 				Value = null;
 				break;
 			case CommonBrushType.Solid:
+			case CommonBrushType.MaterialDesign:
```
ermau: This seems like it adds several changes in behavior to the material type when the brush type is changed, but I don't see a test covering the change. You're changing the value, committing, and then changing the color _again_ via `SetToClosest`.

PR #291, opened 2018-04-24
[build] Move to packagereferences for our nugets

alanmcgovern:
This just moves us to packagereference which is better than the old cruft in all the ways! :p

NOTE: We do force the nugets to restore to a local directory so we have ready/easy access to the `nunit3-console.exe` file so the tests can be executed. If there were a proper nunit-console nuget with  proper msbuild props specifying the path to the executable we wouldn't have to do this. As it stands, this is the only foolproof/safe way to get access to the executable.


alanmcgovern:
If there are any weird issues do ping in the channel. However compiling should now be a simple `msbuild /restore`, or you can do `msbuild /t:Restore && msbuild`

Issue #293, opened 2018-04-30
Mouse scrolling fails in categorized mode


ermau:

ermau:
Appears to be the result of having another "listbox" style item inside it (the group editor).

PR #295, opened 2018-05-01
Collection editor

ermau:
- Add core and windows collection editor:

 - Fixes a crash inside the grouped editor
 - Adds support for having categories auto-expanded

Known issues out of scope:
#294 - Switching between different types can crash
#293 - Mouse wheel scrolling is broken


Xamarin.PropertyEditing/AssignableTypesResult.cs
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		public IReadOnlyCollection<ITypeInfo> AssignableTypes
+		{
+			get;
+		}
+
+		internal IReadOnlyDictionary<IAssemblyInfo, ILookup<string, ITypeInfo>> GetTypeTree ()
+		{
+			var assemblies = new Dictionary<IAssemblyInfo, ILookup<string, ITypeInfo>> ();
```
bleroy: Any particular reason to not use a group by?

Xamarin.PropertyEditing/Reflection/ReflectionObjectEditor.cs
```diff
@@ -138,6 +127,31 @@ public Task<IReadOnlyList<ITypeInfo>> GetAssignableTypesAsync (IPropertyInfo pro
...
+					} else {
+						type = typeof(object);
+					}
+				}
+
+				types = types.Where (t => type.IsAssignableFrom (t));
+
+				return new AssignableTypesResult (types.Select (t => {
+					string asmName = t.Assembly.GetName ().Name;
+					return new TypeInfo (new AssemblyInfo (asmName, isRelevant: asmName.StartsWith ("Xamarin")), t.Namespace, t.Name);
```
bleroy: Should we have a comment in there about why only types in Xamarin assemblies are considered relevant?

Xamarin.PropertyEditing/AssignableTypesResult.cs
ermau: Not off hand, it was already this, I just moved it.

PR #302, opened 2018-05-11
Create new resource core + windows

ermau:


Xamarin.PropertyEditing.Windows/CreateResourceWindow.xaml.cs
```diff
@@ -0,0 +1,54 @@
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Windows;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Windows
+{
+    internal partial class CreateResourceWindow
```
bleroy: You've got some weird mix of tabs and spaces in there.

Xamarin.PropertyEditing/ViewModels/CreateResourceViewModel.cs
```diff
@@ -0,0 +1,395 @@
...
+				OnPropertyChanged (nameof(IsAppliedToAll));
+			}
+		}
+
+		public bool IsAppliedToAll
+		{
+			get { return !this.isKeyed; }
+			set
+			{
+				if (this.isKeyed == !value)
```
bleroy: Can you add a comment about why we have two different properties using the same underlying field? At first, I thought it was a bad copy-paste.

PR #304, opened 2018-05-16
Push fixes to 15.7

ermau:


BretJohnson:
@ermau Can you describe again the bug behavior that this fixes.

FYI - I also just noticed that scrolling via the scroll wheel doesn't seem to work in group mode but does in normal mode. That's probably a separate issue, but may be a nice fix to get into 15.7 as well if it's safe.

ermau:
> FYI - I also just noticed that scrolling via the scroll wheel doesn't seem to work in group mode but does in normal mode.

This is the bug that this fixes.

PR #317, opened 2018-05-18
Entering an out of range color component coerces the value instead of throwing.

### LINKED ISSUE #316: Entering an out of range hue throws, and out of range components trigger validation without coercing values
repro: enter -1 or 361 in the hue field on the HSB or HLS view. Focus out.

expected: the value is coerced into the 0-360 range
actual: exception

Same experience should be available for other components.

bleroy:
… Fixes #316.

### DIFF
--- a/Xamarin.PropertyEditing.Tests/Manual/BrushEditor.md
+++ b/Xamarin.PropertyEditing.Tests/Manual/BrushEditor.md
@@ -193,6 +193,15 @@ Those manual tests should be performed on the standalone app.
 	- The hue picker updates its cursor position.
 	- Other components update as relevant.
 
+## Components are clamped to their range
+
+1. Open the brush editor for the "Mocked Sample 1" , then switch to the HLS view.
+2. For each of the HLS, HSB, RGB, and CYMK, change the values of each field to out of range values and check validation kicks in when moving focus out of the field:
+	- For hue, try -1 and 361. That should cause respectively 0 and 360 to become the values.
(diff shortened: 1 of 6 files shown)


bleroy:
The bug was to make the hue not throw, and behave consistently with other components, but you're right, the other components are also not behaving like the WPF color editor (which clamps all components in their respective range). I'll modify #316 and this PR accordingly.

Xamarin.PropertyEditing.Windows/ByteToPercentageConverter.cs
```diff
@@ -10,21 +10,14 @@ namespace Xamarin.PropertyEditing.Windows
...
-			if (value is double doubleValue) {
-				if ((doubleValue < 0) || (doubleValue > 100)) return DependencyProperty.UnsetValue;
-				return System.Convert.ToByte (doubleValue * 2.55d);
-			}
-			return DependencyProperty.UnsetValue;
-		}
+			=> !(value is double doubleValue) ? DependencyProperty.UnsetValue
+				: doubleValue < 0 ? 0
+				: doubleValue > 100 ? 255
+				: System.Convert.ToByte (doubleValue * 2.55d);
```
ermau: I find this being concise at the cost of readability.

bleroy: I don't know about that, I happen to find it more readable in fact: you clearly see each case and its outcome.

PR #321, opened 2018-06-12
[Win] Fix double click on scrollbar selecting

ermau:


Xamarin.PropertyEditing.Windows/ResourceSelectorWindow.xaml.cs
```diff
@@ -50,6 +52,17 @@ private void OnListDoubleClick (object sender, MouseButtonEventArgs e)
...
+			Point pos = e.GetPosition (this.list);
+			var element = this.list.InputHitTest (pos) as DependencyObject;
+			while (element != null) {
+				if (element is ListBoxItem)
+					break;
+				if (element is ListBox)
+					return;
+
+				element = VisualTreeHelper.GetParent (element);
+			}
```
bleroy: So if you never find an item or a listbox, and element gets null, `DialogResult` still gets set to true? Is that the intention?

ermau: Shouldn't be possible, but addressed anyway to be safe, good catch.

PR #329, opened 2018-06-27
Improve color precision, update test plan, prevent unchanged component values from changing the brush

bleroy:


bleroy:
Needs a little more work.

Issue #131, opened 2018-01-23
Mac resource window


ermau:
- [x] Core - #129
 - [x] Design
 - [x] Localized
 - Behavior:
   - [ ] Keyboard control
   - [ ] Accessibility
   - [ ] Selection
 - Theming:
   - [x] Light themed
   - [x] Dark themed
   - [x] High contrast

vancura:
## Standard Property Pad with a tooltip

<img width="461" alt="mac resource selector 2x" src="https://user-images.githubusercontent.com/4982/35686245-1fa15d3e-076c-11e8-8237-1c98a0c77371.png">

## Selecting a color resource

- The headline in the popover contains _Color_ – that's when it's clear that only one type of the resource could be selected (color in this case).
- There's the _all/local/shared_ switch in the footer, which just filters the list to see the items based on the selection.
- Shared items are marked with an icon (also seen in the footer next to the _shared_ button)
- There's also a _Preview_ icon to toggle the preview pane (see below). Here it's not enabled, the icon is not highlighted.

<img width="628" alt="mac resource selector popover colors 2x" src="https://user-images.githubusercontent.com/4982/35686446-9227a08e-076c-11e8-9b46-f563b1ca83b5.png">

## Color resource is selected

- The headline in the popover still displays _Color_, since we don't allow to select anything but a color.
- The color resource is selected, so we need to mark the item in the table as selected (following Apple HIG, with white text and `~sel` icon, also with a white stroke around the color).
- The color is set in the property in the property pad.
- The user can now click anywhere, the popover hides and the new value remains in the property pad.

<img width="628" alt="mac resource selector popover colors selected 2x" src="https://user-images.githubusercontent.com/4982/35686590-e7668a9c-076c-11e8-8d54-fbcd6af4027e.png">

## Selecting a String resource

- The headline in the popover displays _String_ since we don't allow to select anything but a String.

<img width="628" alt="mac resource selector popover generic 2x" src="https://user-images.githubusercontent.com/4982/35689753-7c5d53cc-0774-11e8-9b72-5413c6cc53db.png">

## Selecting a generic resource

Since we allow to select one of many types of resources, we do a couple of things:

- The headline in the popover doesn't display any explicit resource type.
- The table contains a type rectangle too, so it's obvious what's the resource type.

[...]

PR #334, opened 2018-07-11
[Mac] Initial Request Resource Implementation

CartBlanche:
Uses Larry's Brush Views.

Linked to - https://github.com/xamarin/Xamarin.PropertyEditing/issues/131


Xamarin.PropertyEditing.Mac/Controls/RequestResourceView.cs
```diff
@@ -0,0 +1,507 @@
...
+				throw new ArgumentNullException (nameof (propertyViewModel));
+
+			this.viewModel = propertyViewModel;
+		}
+
+		public ResourceSelectorViewModel ViewModel => this.viewModel;
+
+		public override nint GetRowCount (NSTableView tableView)
+		{
+			return this.viewModel.Resources.Count ();
```
ermau: We can convert this to the property when it's based on #339

ermau:
Additionally if the value is already set to a resource, it should show as selected when you open the selector.

CartBlanche:
@ermau I'll double check, but so far the filtering should be handled by the VM
`resourceSelectorPanel.ViewModel.FilterText = searchResources.Cell.Title;`
I'll see if something else is being corrupted then that property is being set.

ermau:
It is filtering, it's just showing the wrong results, the first place I'd be looking is to make sure the delegates are getting the right data when they look it up after a reload. It's working on Windows.

CartBlanche:
UI changes from @vancura 
- [ ] Make it Standard selection blue with white text? http://vncr.in/tNbW
- [ ] This icon is blurry: http://vncr.in/tNuI
- [ ] This would be nice to pad from top AND bottom the same way: http://vncr.in/tNqs
- [ ] Not centered: http://vncr.in/tNij

Xamarin.PropertyEditing.Mac/Controls/RequestResourceView.cs
```diff
@@ -0,0 +1,507 @@
...
+			previewView = new NSView (new CGRect (20, 0, frame.Width - 30, frame.Height));
+			previewView.Hidden = true; // Hidden until a resource is selected and a preview is available for it.
+			AddSubview (previewView);
+		}
+
+		NSView GetPreviewView (Type type)
+		{
+			Type[] genericArgs = null;
+			Type previewRenderType;
+			if (!ResourceTableDelegate.ValueTypes.TryGetValue (type, out previewRenderType)) {
```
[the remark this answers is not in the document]
CartBlanche: Done.

Xamarin.PropertyEditing.Mac/Controls/RequestResourceView.cs
```diff
@@ -0,0 +1,507 @@
...
+
+			case RequestResourcePanel.ResourceNameColId:
+				var nameView = (NSTextField)tableView.MakeView (nameIdentifier, this);
+				if (nameView == null) {
+					nameView = CreateReadOnlyTextField (resource.Name, nameIdentifier);
+				}
+				return nameView;
+
+			case RequestResourcePanel.ResourceValueColId:
+				var valueView = MakeValueView (resource.RepresentationType.Name + valueIdentifier, resource.RepresentationType, tableView);
```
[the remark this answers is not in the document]
CartBlanche: Fixed.

Xamarin.PropertyEditing.Mac/Controls/RequestResourceView.cs
```diff
@@ -0,0 +1,507 @@
...
+		internal const string ResourceTypeColId = "ResourceType";
+		internal const string ResourceNameColId = "ResourceName";
+		internal const string ResourceValueColId = "ResourceValue";
+
+		private NSTableView resourceTable;
+		private ResourceTableDataSource dataSource;
+
+		private ResourceSelectorViewModel viewModel;
+		public ResourceSelectorViewModel ViewModel => this.viewModel;
+		public Resource SelectedResource => this.viewModel.Resources.ElementAt ((int)this.resourceTable.SelectedRow) as Resource;
```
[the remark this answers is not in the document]
CartBlanche: Thanks!

Xamarin.PropertyEditing.Mac/Controls/RequestResourceView.cs
```diff
@@ -0,0 +1,507 @@
...
+
+			ReloadData ();
+		}
+
+		internal void ReloadData ()
+		{
+			this.resourceTable.ReloadData ();
+		}
+
+		internal void RepositionControls (bool showPreview)
```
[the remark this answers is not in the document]
CartBlanche: Removed.

Xamarin.PropertyEditing.Mac/Controls/RequestResourceView.cs
CartBlanche: Done.

CartBlanche:
@ermau I believe I've responded, either code-wise or text-wise, to your requested changes.
@vancura Is back from holiday on Monday.

Xamarin.PropertyEditing.Mac/Controls/BrushEditorControl.cs
```diff
@@ -112,8 +112,7 @@ string GetTitle ()
 					break;
 			}
 
-			//return ViewModel.Resource == null ? title : $"{title} - (Resource: {ViewModel?.Resource?.Name})";
-			return title;
+			return ViewModel.Resource == null ? title : $"{title} - (Resource: {ViewModel?.Resource?.Name})";
```
ermau: Resource needs to be localized.

CartBlanche: Doh! I thought I'd caught all of these.

CartBlanche:
@vancura's new icons are now part of this PR.

Xamarin.PropertyEditing.Mac/Controls/RequestResource/ResourceTableDelegate.cs
```diff
@@ -0,0 +1,171 @@
...
+			if (notification.Object is NSTableView tableView) {
+				NSImageView iconColumn;
+				if (previousRow != -1) {
+					iconColumn = tableView.GetView (0, previousRow, false) as NSImageView;
+					iconColumn.Image = PropertyEditorPanel.ThemeManager.GetImageForTheme ("resource-editor-32");
+				}
+
+				if (tableView.SelectedRow != -1) {
+					iconColumn = tableView.GetView (0, tableView.SelectedRow, false) as NSImageView;
+					iconColumn.Image = PropertyEditorPanel.ThemeManager.GetImageForTheme ("resource-editor-32", "~sel");
```
ermau: This throws NRE if you're opening the resource window for a property which already has a resource selected.

ermau:
A property which is set to null (reset value) doesn't have its resource set when selecting them.

Xamarin.PropertyEditing.Mac/Controls/RequestResource/ResourceTableDelegate.cs
CartBlanche: Fixed.

CartBlanche:
> A property which is set to null (reset value) doesn't have its resource set when selecting them.
@ermau that behaviour appears to be consistent on both Windows and Mac. Only setting Solid colour first (after a Reset) then allows you to select a Palette or Resource. I've not dug further to find out why, but could this be an issue in the VM?

> The selected row highlight disappears when the item is filtered out and back in, despite still being the selected resource.
This also seems to be consistent across Windows and Mac, unless I've misunderstood what you've written. If this requires a change to both platforms, I'd prefer do that in a followup PR.

ermau:
> @ermau that behaviour appears to be consistent on both Windows and Mac. Only setting Solid colour first (after a Reset) then allows you to select a Palette or Resource. I've not dug further to find out why, but could this be an issue in the VM?

Entirely possible. Might also be a bug in the mock backing.

> This also seems to be consistent across Windows and Mac, unless I've misunderstood what you've written. If this requires a change to both platforms, I'd prefer do that in a followup PR.

This is likely a bug that exists on each platform individually, just worry about the Mac side here.

CartBlanche:
> A property which is set to null (reset value) doesn't have its resource set when selecting them.

Eric fixed this with this PR...
https://github.com/xamarin/Xamarin.PropertyEditing/pull/335

CartBlanche:
@ermau Looks like the "unfocused" highlighting issue is the default in Dark mode. I finally managed to learn enough Objective C to put an Xcode version of theming. See screenshot.

screen shot 2018-08-29 at 14 17 38

Both our standalone and the Objective C version exhibit exactly the same issue.

As this is an aesthetic issue, I'd suggest this doesn't hold up this PR and we can investigate our own workout for our various ListViews( maybe having custom NSOutlineView and NSTableView controls that render correctly)

ermau:
- Values should not be changing as I select the different resource, it's only on commit.
- Double clicking a resource should commit (set and close the window) but does nothing.
- Entering a filter, closing the dialog and reopening it has it still filtered, but the entered filter text is blank. It should not maintain the filter.

CartBlanche:
> Values should not be changing as I select the different resource, it's only on commit.

@ermau That may be how it works on Windows, but in XCode the colour picker updates immediately on selection of the colour.  I can post a video if you'd like to see an example of this when related to changing the text colour of a control, for example

CartBlanche:
> Entering a filter, closing the dialog and reopening it has it still filtered, but the entered filter text is blank. It should not maintain the filter.

@ermau Fixed. Filter no longer maintained.

CartBlanche:
> Double clicking a resource should commit (set and close the window) but does nothing.

Even though double clicking isn't standard Mac behaviour it is a slight better UX if you do double click and it auto-closes the popup. So I've added that functionality in.

ermau:
> @ermau That may be how it works on Windows, but in XCode the colour picker updates immediately on selection of the colour. I can post a video if you'd like to see an example of this when related to changing the text colour of a control, for example

We are not trying to copy Xcode. Users should be able to expect a consistent experience for when their values actually change across the product.

CartBlanche:
> We are not trying to copy Xcode. Users should be able to expect a consistent experience for when their values actually change across the product.

It isn't just Xcode. MacOS generally works that way. 

I think users will expect the values to change to be consistent with the Operating System they are using, not having tonnes of Windowisms in what is supposed to be a *native* Mac app.

Not even the Mac preferences dialog has an OK button. Values are auto-commited. It is standard MacOS behaviour.

screen shot 2018-09-04 at 18 58 21

My double click change above doesn't change the auto-commit that is a natural part of Mac, it simply closes the pop-up after the auto-commit happens.

If you feel really strongly about it let's chat to @vancura and @alanmcgovern about redesigning both the Resource pop-up and the Brush Editor pop-up to include an OK button, if we want to make it like Windows. Without an OK button it doesn't make sense to have *extra* clicks.

ermau:
I can see this two ways. One is as you've said, auto-commit may feel more mac-ish. Although I wouldn't use the preferences screen as an example. The one inside the app we actually live in does _not_ autocommit and has ok/cancel buttons.

On the other hand it's possible to provide a _better_ experience by separating preview from commit. We have this preview section in the selector window, but it seems rather redundant for some properties that may see the change immediately visible on the designer surface.

I'm going to approve and merge this, but we'll discuss further internally.

CartBlanche:
OK.

PR #335, opened 2018-07-12
[build] Remove touchdown from the normal restore path

alanmcgovern:
Touchdown is only needed for localization and we only run that rarely,
so remove that source from the standard NuGet.Config so we don't have
to query it as part of our normal build/restore process.


netonjm:
Good idea!

PR #337, opened 2018-07-13
Add source descriptor and rebase object VM

ermau:


Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
```diff
@@ -69,7 +69,7 @@ public string TypeName
 
 		public event EventHandler<EditorPropertyChangedEventArgs> PropertyChanged;
 
-		public Func<IPropertyInfo, object, object> ValueEvaluator
+		public Func<IPropertyInfo, object, object, object> ValueEvaluator
```
bleroy: I think a doc comment is absolutely necessary here, as the object types obfuscate the semantics.

Xamarin.PropertyEditing/ValueInfo.cs
```diff
@@ -15,7 +12,7 @@ public T Value
 		}
 
 		/// <summary>
-		/// Gets or sets a descriptor of the value source, such as a resource reference or binding description.
+		/// Gets or sets a descriptor of the value, such as a string value name or <see cref="ITypeInfo"/>.
```
bleroy: This is actually difficult and subtle. For developers integrating the property panel, adding more details about the difference between value and source descriptors would be very useful.

Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
ermau: It's only tests, but I can add something.

Xamarin.PropertyEditing/ValueInfo.cs
ermau: What more can I add than the exact types you might expect in each property like I did?

bleroy: Two things I think: first, some explanation about why they are separated (we had to debate whether they should, which indicates it's not a trivial thing), and also maybe more details about how you'd implement it. When integrating this, you'll have to set those values, but you'll also have to use them, unless I misunderstood. Maybe walking through the top-level steps of a use case would be useful.

ermau: Discussion details are in the commit expanded message. We didn't really debate whether they should be separated, just how to separate them, the need for separation was obvious.

I can add a seealso targeted at `IObjectEditor.SetValueAsync`, implementation details are provided there including examples of this property. I'll also see if I need to add more detail there.

ermau:
@bleroy added extra docs.

Xamarin.PropertyEditing/IObjectEditor.cs
```diff
@@ -112,6 +116,13 @@ public interface IObjectEditor
 		/// whereever possible.</para>
 		/// <para>If the platform can know the value of a property when unset, it should return that value and the <see cref="ValueSource.Default"/>
 		/// source. If the platform only knows that the value is unset, use <see cref="ValueSource.Unset"/> instead.</para>
+		///<para>When the property's value <see cref="ValueInfo{T}.Source"/> is <see cref="ValueSource.Resource"/>,
+		/// the <see cref="ValueInfo{T}.SourceDescriptor"/> should be set to a <see cref="Resource"/> instance. Implementers should
+		/// strive to retrieve the resource's value and use <see cref="Resource{T}"/> isntead. For a source of
```
bleroy: You have a typo on "isntead".

PR #356, opened 2018-07-26
[Mac] Fix tab order

CartBlanche:
Extracted what's needed from CommonRatio to fix the tab ordering in master.


ermau:
This is not at all extracted, there's a ton of unrelated changes.

PR #360, opened 2018-08-06
[Mac] Add GetImageForTheme call to simplify theme naming calls.

CartBlanche:


Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
```diff
@@ -27,5 +29,23 @@ protected override void SetTheme ()
...
+		public string GetImageNameForTheme (string imageNamed, string selectedSuffix = "")
+		{
+			return (Theme == PropertyEditorTheme.Dark ? imageNamed + "~dark" : imageNamed) + selectedSuffix;
+		}
+
+		public NSImage GetImageForTheme (string imageNamed, string selectedSuffix = "")
+		{
+			var imageNameForTheme = GetImageNameForTheme (imageNamed, selectedSuffix);
+			NSImage themeImage;
+			if (this.themeCache.TryGetValue (imageNameForTheme, out themeImage)) {
```
CartBlanche: @ermau I was contemplating wrapping this into something more async friendly. Thoughts?

ermau: I don't think we want to complicate every image load with asyncifying it, especially since we'll only have the delay once.

ermau: You can simplify this by doing something like:

```csharp
if (!this.themeCache.TryGetValue (imageNameForTheme, out NSImage themeImage)) {
    themeImage = NSImage.ImageNamed (imageNameForTheme);
    this.themeCache[imageNameForTheme] = themeImage;
}
return themeImage;
```

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
```diff
@@ -27,5 +29,23 @@ protected override void SetTheme ()
 					break;
 			}
 		}
+
+		public string GetImageNameForTheme (string imageNamed, string selectedSuffix = "")
```
ermau: What's the suffix for?

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
```diff
@@ -1,12 +1,14 @@
...
 using System.Drawing;
+using System.Collections.Generic;
 using AppKit;
 
 namespace Xamarin.PropertyEditing.Themes
 {
 	public class MacThemeManager : BaseThemeManager
 	{
 		public NSAppearance CurrentAppearance { get; private set; }
+		private Dictionary<string, NSImage> themeCache = new Dictionary<string, NSImage> ();
```
ermau: Might be worth clearing this when the theme changes.

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
CartBlanche: Sure.

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
CartBlanche: Possibly. If the user changes between themes we'd be taking the hit again on adding the items to the new list. Without the clear, we'd just keep adding and right now we only have 2 possible themes. So do we want to take that hit or can we live with the extra memory it would take up, holding the images in memory for 2 themes?

ermau: The hit is relatively minor, we're not even caching currently. In practice a user isn't going to be switching themes constantly, it's not worth wasting a bunch of memory over.

Xamarin.PropertyEditing.Mac/Themes/MacThemeManager.cs
```diff
@@ -26,6 +28,24 @@ protected override void SetTheme ()
...
+		public string GetImageNameForTheme (string imageNamed, string selectedSuffix = "")
+		{
+			return (Theme == PropertyEditorTheme.Dark ? imageNamed + "~dark" : imageNamed) + selectedSuffix;
+		}
+
+		public NSImage GetImageForTheme (string imageNamed, string selectedSuffix = "")
+		{
+			var imageNameForTheme = GetImageNameForTheme (imageNamed, selectedSuffix);
+
+			if (!this.themeCache.TryGetValue (imageNameForTheme, out NSImage themeImage)) {
```
CartBlanche: I'm only just discovering how good these inline declarations are and weirdly though the scope would only be within the if. But since it is an `out`, it make sense.

CartBlanche:
@ermau changes made. Worth pulling in?

PR #366, opened 2018-08-17
Fix resource issues

ermau:
This reverts an invalid change with setting resources and fixes an issue from the source/value descriptor break out when setting resource values.


CartBlanche:
LGTM and works as advertised!

PR #368, opened 2018-08-23
Fix "add ___" popup cancellation combobox results

ermau:


CartBlanche:
I don't think this is related to this PR, but thought I'd mention it and can create a separate issue.  Basically If I add more than 1 Object to the Collection, it crashes, with attached error.

screen shot 2018-08-25 at 19 44 43

ermau:
> Basically If I add more than 1 Object to the Collection, it crashes, with attached error.

This is probably #294

Issue #376, opened 2018-08-30
Windows Basic autocomplete


ermau:
The simple replace all implementation for autocomplete (#375)

Issue #375, opened 2018-08-30
Basic autocomplete core


ermau:
There's two tiers of autocomplete we can support. One is a simple replacement for things like resource names as Android has today, and the other is closer to what you'd expect out of an IDE for things like markup extensions. This is the issue for the former, the later will live in #307 

![image](https://user-images.githubusercontent.com/156582/40138566-17703602-591b-11e8-9b82-8cb1608c9d09.png)

CartBlanche:
Should this issue be closed, shouldn't it stay open until the Mac side is completed?

PR #379, opened 2018-08-31
Basic autocomplete support

ermau:
Completes #375 
Completes #376

### DIFF
--- a/Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
+++ b/Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using Xamarin.PropertyEditing.Reflection;
 using Xamarin.PropertyEditing.Tests.MockControls;
@@ -29,7 +30,7 @@ namespace Xamarin.PropertyEditing.Tests
 	}
 
 	internal class MockObjectEditor
-		: IObjectEditor, IObjectEventEditor
+		: IObjectEditor, IObjectEventEditor, ICompleteValues
 	{
 		public MockObjectEditor ()
 		{
@@ -106,6 +107,12 @@ namespace Xamarin.PropertyEditing.Tests
 		}
 
 		public IReadOnlyList<IObjectEditor> DirectChildren
+		{
+			get;
+			set;
+		}
+
+		public IResourceProvider Resources
 		{
 			get;
 			set;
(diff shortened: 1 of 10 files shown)


Xamarin.PropertyEditing.Windows/EntryPopup.cs
```diff
@@ -9,6 +8,18 @@ namespace Xamarin.PropertyEditing.Windows
 	internal class EntryPopup
 		: Popup
     {
+		static EntryPopup ()
+		{
+			var existing = IsOpenProperty.GetMetadata (typeof(Popup));
```
bleroy: You might want to add a comment explaining this.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -77,6 +78,32 @@ public override Resource Resource
...
+					this.autocomplete = null;
+					this.autocompleteCancel?.Cancel();
+					this.autocompleteCancel = null;
+				}
+			}
+		}
+
+		public IReadOnlyList<string> AutocompleteItems => this.autocomplete;
+
+		public string PreviewCustomExpression
```
bleroy: Write-only property is weird. Why not just call `UpdateAutoComplete` directly instead?

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -359,6 +407,49 @@ private async void OnCreateBinding ()
...
+					tasks.Remove (results);
+
+					if (list == null) {
+						list = results.Result;
+						common = new HashSet<string> (list);
+					} else
+						common.IntersectWith (results.Result);
+				} while (tasks.Count > 0 && !cancel.IsCancellationRequested);
+
+				this.autocomplete.Reset (list.Where (common.Contains));
```
bleroy: It looks like when this is called, the autocomplete will be the intersection of the first list returned by any of the editors that support auto-complete, and of the intersection of everything each of them returned. It actually looks like this would be the same as common, except for order. What exactly is the expected behavior when there are multiple auto-completing editors present? Seems like the way it's implemented, it's items that are common to all editors, in the order returned by whichever finished first. If that is the expected behavior, I'd add a comment confirming it, and why that is.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
ermau: Binding-ease.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
ermau: > Seems like the way it's implemented, it's items that are common to all editors, in the order returned by whichever finished first.

Correct, I'll add a comment.

Xamarin.PropertyEditing.Tests/MockObjectEditor.cs
```diff
@@ -332,6 +339,21 @@ public Task<ITypeInfo> GetValueTypeAsync (IPropertyInfo property, PropertyVariat
...
+		}
+
+		public async Task<IReadOnlyList<string>> GetCompletionsAsync (IPropertyInfo property, string input, CancellationToken cancellationToken)
+		{
+			if (Resources == null)
+				return Array.Empty<string> ();
+
+			input = input.Trim ().TrimStart('@').ToLower ();
+			var resources = await Resources.GetResourcesAsync (Target, property, cancellationToken);
+			return resources.Where (r => r.Name.ToLower().Contains (input) && r.Name.Length > input.Length).Select (r => "@" + r.Name).ToList ();
```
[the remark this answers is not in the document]
ermau: Yes, probably.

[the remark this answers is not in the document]
ermau: Correct.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
ermau: Comments addressing behavior are on the API

CartBlanche:
@ermau I assume there will be a Mac side to this?

CartBlanche:
👍

CartBlanche:
Once the CustomExpression is selected. When you re-open the CustomExpression Dialog, the previously selected value isn't displayed in the CustomExpression screen.

PR #389, opened 2018-09-05
[Mac] Initial implemention of Thickness Editor.

CartBlanche:
I've made this behave similar to how the Rectangle Editor works, so as such it does not support any images in this version to replace the  label text, just as the Point Editor doesn't support images right now. So it is consistent with what we currently have. The idea being that adding images would require asking Vaclav for extra art. Hope to come back to this closer to release to add the images into this and the Point editor.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/ThicknessEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/ThicknessEditorControl.cs
@@ -1,0 +1,72 @@
+using System;
+using AppKit;
+using CoreGraphics;
+using Xamarin.PropertyEditing.Drawing;
+using Xamarin.PropertyEditing.Mac.Resources;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class CommonThicknessEditorControl : BaseRectangleEditorControl<CommonThickness>
+	{
+		public CommonThicknessEditorControl ()
+		{
+			XLabel.Frame = new CGRect (28, 27, 50, 22);
+			XLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize);
+			XLabel.StringValue = "LEFT";
+
+			XEditor.Frame = new CGRect (4, 46, 90, 20);
+
+			YLabel.Frame = new CGRect (160, 27, 45, 22);
+			YLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize);
+			YLabel.StringValue = "TOP";
+
+			YEditor.Frame = new CGRect (132, 46, 90, 20);
+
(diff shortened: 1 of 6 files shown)


Xamarin.PropertyEditing.Mac/Controls/ThicknessEditorControl.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using AppKit;
+using CoreGraphics;
+using Xamarin.PropertyEditing.Drawing;
+using Xamarin.PropertyEditing.Mac.Resources;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal abstract class ThicknessEditorControl<T> : BaseRectangleEditorControl<T>
```
ermau: There's only one thickness type, `CommonThickness`. We don't need any generics here.

Xamarin.PropertyEditing.Mac/Controls/ThicknessEditorControl.cs
```diff
@@ -0,0 +1,76 @@
...
+	{
+		public ThicknessEditorControl ()
+		{
+			// TODO localize
+			XLabel.Frame = new CGRect (28, 27, 50, 22);
+			XLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize);
+			XLabel.StringValue = "LEFT";
```
ermau: Why would we have a todo for localization when we can do it now?

CartBlanche: The comment was copy and pasted in and for all the those types of controls, we never Localised, X, Y, Width or Height as the IDEs don't localise those, they are kept in English, IIRC after chatting to @alanmcgovern. Happy to localise all these if I've misunderstood what needed to be localised and what didn't.

Xamarin.PropertyEditing.Mac/Controls/ThicknessEditorControl.cs
CartBlanche: That's across all platforms? Should we remove the generics for Point, Rectangle and Size too, down the line?

CartBlanche:
The Rectangle Editor was done at about the same time as the Point Editor, IIRC. I worked with @vancura on it, back then.

ermau:
We'll have to add something extra, fixed text labels aren't really platform or even property agnostic.

CartBlanche:
@ermau shouldn't we then do a pass on Point, Size, Rectangle as well then, as they are all fixed text labels too?

CartBlanche:
@ermau @alanmcgovern @vancura Has there been a final decision on whether we are changing the labels to images on Mac? Xcode uses text.

vancura:
There was no discussion AFAIK.

Are we talking just about Point / Size / Rectangle? Anything else?

ermau:
I'll move the discussion to an issue because it's a bit broader.

Issue #377, opened 2018-08-30
Mac basic autocomplete


ermau:
The simple replace all implementation for autocomplete (#375)

PR #394, opened 2018-09-11
[Mac] Initial implementation of Autocompletion.

CartBlanche:
Closes - https://github.com/xamarin/Xamarin.PropertyEditing/issues/377

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
@@ -1,0 +1,132 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+using System.Reflection;
+using AppKit;
+using Foundation;
+using ObjCRuntime;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac.Controls
+{
+	internal class AutocompleteComboBox : NSComboBox
+	{
+		private readonly PropertyViewModel viewModel;
+		private readonly PropertyInfo previewCustomExpressionPropertyInfo;
+		public PropertyInfo PreviewCustomExpressionPropertyInfo
+		{
+			get { return this.PreviewCustomExpressionPropertyInfo; }
+			set {
+				this.PreviewCustomExpressionPropertyInfo = value;
+			}
+		}
+		private ObservableCollectionEx<string> values;
+
+		protected IHostResourceProvider HostResources {
+			get;
+			private set;
+		}
+
+		public AutocompleteComboBox (IHostResourceProvider hostResources, PropertyViewModel viewModel, ObservableCollectionEx<string> values, PropertyInfo previewCustomExpressionPropertyInfo)
+		{
+			if (hostResources == null)
+				throw new ArgumentNullException (nameof (hostResources));
+
+			if (viewModel == null)
+				throw new ArgumentNullException (nameof (viewModel));
+
+			if (values == null)
+				throw new ArgumentNullException (nameof (values));
+
+			if (previewCustomExpressionPropertyInfo == null)
+				throw new ArgumentNullException (nameof (previewCustomExpressionPropertyInfo));
+
+			HostResources = hostResources;
+
+			this.viewModel = viewModel;
+			this.values = values;
+			this.previewCustomExpressionPropertyInfo = previewCustomExpressionPropertyInfo;
+
+			this.values.CollectionChanged += (sender, e) => {
+				switch (e.Action) {
+					case NotifyCollectionChangedAction.Add:
+						if (e.NewStartingIndex == -1 || e.NewItems == null) {
+							Reset ();
+						} else {
+							var items = e.NewItems;
+							int startIndex = e.NewStartingIndex;
+							if (startIndex != -1 && startIndex < items.Count) {
+								for (var i = 0; i < items.Count; i++) {
+									Insert (new NSString ((string)items[i]), startIndex);
+									startIndex++;
+								}
+							} else {
+								PopulateComboBoxItems (items);
+							}
+						}
+					break;
+
+					case NotifyCollectionChangedAction.Remove:
+						if (e.OldStartingIndex == -1 || e.OldItems == null) {
+							Reset ();
+						} else {
+							var items = e.OldItems;
+							int startIndex = e.OldStartingIndex;
+							if (startIndex != -1 && startIndex < items.Count) {
+								for (var i = 0; i < items.Count; i++) {
+									RemoveAt (startIndex);
+								}
+							}
+						}
+					break;
+
+				case NotifyCollectionChangedAction.Move:
+				case NotifyCollectionChangedAction.Replace:
+				case NotifyCollectionChangedAction.Reset:
+					Reset ();
+					break;
+
+				default:
+						break;
+				}
+			};
+
+			// Maximum number of items in the drop-down before scrolling kicks in.
+			VisibleItems = 8;
+
+			var selectorPopup = new Selector ("popUp:");
+
+			Changed += (sender, e) => {
+				if (!Cell.AccessibilityExpanded) {
+					Cell.PerformSelector (selectorPopup);
+				}
+
+				UpdatePropertyInfo ();
+			};
+
+			UpdatePropertyInfo ();
+
+			ViewDidChangeEffectiveAppearance ();
+		}
+
+		private void Reset ()
+		{
+			RemoveAll ();
+			PopulateComboBoxItems (this.values);
+		}
+
+		private void PopulateComboBoxItems (IList items)
+		{
+			for (var i = 0; i < items.Count; i++) {
+				Add (new NSString (this.values[i]));
+			}
+		}
+
+		public void UpdatePropertyInfo ()
+		{
+			this.previewCustomExpressionPropertyInfo.SetValue (this.viewModel, StringValue);
+		}
+	}
+}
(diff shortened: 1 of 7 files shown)


Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteTextField.cs
```diff
@@ -0,0 +1,75 @@
...
+		{
+			var values = customcompleteItemsPropertyInfo.GetValue (this.viewModel);
+			if (values != null && this.autocompleteValues == null)
+				this.autocompleteValues = ((IReadOnlyList<string>)values);
+
+			return autocompleteValues.ToArray ();
+		}
+	}
+
+	internal class AutocompleteTextFieldDelegate : NSTextFieldDelegate
```
ermau: This could be a child class to `AutocompleteTextField`. Nothing else will need to interact with it, and then `Suggestions()` (which should be `GetSuggestions()`) can be private.

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteTextField.cs
```diff
@@ -0,0 +1,75 @@
...
+			Delegate = new AutocompleteTextFieldDelegate ();
+		}
+
+		internal string[] Suggestions ()
+		{
+			var values = customcompleteItemsPropertyInfo.GetValue (this.viewModel);
+			if (values != null && this.autocompleteValues == null)
+				this.autocompleteValues = ((IReadOnlyList<string>)values);
+
+			return autocompleteValues.ToArray ();
```
ermau: If `values` is `null`, which it can be if a `PreviewCustomExpression` has not been set, this can NRE.

CartBlanche:
Video of Autocomplete in action - https://1drv.ms/v/s!Agbp4B3CE3TVgpxIMNi02wn_DB8lVA

ermau:
This doesn't re-query the backend provider at all on changes, the VM is responsible for the filtering, not the UI. If the provider didn't provide any results in this example until you added "@", no results will be picked up. Additionally can we hide the drop down button? This is not meant to be a combobox.

CartBlanche:
Example of Autocompletion with delayed insertions on startup...
autocomplete2-1

Example of Autocompletion with delayed insertions while filtering, after the list has loaded...
autocomplete2-2

We could certainly support it, but IMO it makes for a horrendous UX. I still think we have to ensure the list is full loaded before it returns. Maybe `PreviewItems` fires an ListLoaded event, that we hook into, then blit to the combobox/or other list. So the control would say **Loading** and be readonly until we are notified that we have the full list.

CartBlanche:
@ermau I switched away from the NSTextField because trying to get the auto-complete working wasn't as straight forward as initially thought and it had the weird side effects we saw when I initially put up the PR, due to us needing to road-trip to the VM. As we are essentially providing a list of items to pick from, while also allowing the user to enter a custom expression the NSComboBox, with its build in list handling, seems an appropriate compromise as users know how combo boxes are expected to work. I've had a discussion with @vancura, as to whether from a design perspective, if he thinks it breaks the aesthetic or UX and he doesn't think it does either.

ermau:
> We could certainly support it, but IMO it makes for a horrendous UX. I still think we have to ensure the list is full loaded before it returns. 

In the 99% case this is essentially what happens.

> As we are essentially providing a list of items to pick from, while also allowing the user to enter a custom expression the NSComboBox

It's not a fixed list though like a combo box, our usage is expectation defying, especially when we get to sub-input replacement instead of the whole thing like we're doing now. We can take this for now but it just means we're probably going to have to replace it later.

CartBlanche:
> sub-input replacement

Could you please clarify??

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
```diff
@@ -0,0 +1,85 @@
...
+				throw new ArgumentNullException (nameof (values));
+
+			if (previewCustomExpressionPropertyInfo == null)
+				throw new ArgumentNullException (nameof (previewCustomExpressionPropertyInfo));
+
+			this.viewModel = viewModel;
+			this.values = values;
+			this.previewCustomExpressionPropertyInfo = previewCustomExpressionPropertyInfo;
+
+			this.values.CollectionChanged += (sender, e) => {
```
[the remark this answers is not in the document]
ermau: You're missing handling indexes, they may not be appended at the end.

CartBlanche:
@ermau Just a reminder there was also this PR.

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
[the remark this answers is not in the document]
CartBlanche: @ermau so would that be using the `e.NewStartIndex` and `e.OldStartIndex` during the `.Add` and `.Remove` actions, respectively? We don't seem to be adding indexes anywhere else with these `ObservableCollectionEx`.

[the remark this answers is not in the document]
ermau: > so would that be using the e.NewStartIndex and e.OldStartIndex during the .Add and .Remove actions, respectively?

Yes.

> We don't seem to be adding indexes anywhere else with these ObservableCollectionEx.

I don't follow. `ObservableCollection` provides indexes for add and remove, `Ex` is a subclass.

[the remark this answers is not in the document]
ermau: > We don't seem to be handling indexes anywhere else with these ObservableCollectionEx.

Most "handling" cases in the code base are actually done by WPF, which does.

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
```diff
@@ -0,0 +1,98 @@
...
+				switch (e.Action) {
+					case NotifyCollectionChangedAction.Add:
+					if (e.NewItems != null) {
+						var items = (ObservableCollectionEx<string>)e.NewItems;
```
ermau: There's no guarantee this will be true, nor is it even needed.

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
```diff
@@ -0,0 +1,98 @@
...
+					break;
+
+					case NotifyCollectionChangedAction.Remove:
+						if (e.OldItems != null) {
+							var items = (ObservableCollectionEx<string>)e.OldItems;
+							int startIndex = e.OldStartingIndex;
+							if (startIndex != -1 && startIndex < items.Count) {
+								for (var i = 0; i < items.Count; i++) {
+									RemoveAt (startIndex);
+									startIndex++;
```
ermau: `0 | 1 | 2 | 3 | 4 | 5`
```
startIndex = 1; count = 2;
RemoveAt (startIndex++);
```
This leaves you in a state where:
```
0 | 2 | 3 | 4 | 5
startIndex == 2;
RemoveAt (startIndex++);
```
You've removed item 1 and 3 instead of 1 and 2. Don't increment the index when removing by index.

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
```diff
@@ -0,0 +1,98 @@
...
+								}
+							} else {
+								PopulateComboBoxItems (items);
```
ermau: Why are you adding items that are meant to be removed?

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
CartBlanche: OK Sure. Also just noticed I might end up with an out of bounds on the **RemoveAt** too.
I also noticed that everywhere else, where we are tracking **NotifyCollectionChangedAction.Add** and **NotifyCollectionChangedAction.Remove**, that we aren't using **NewStartingIndex** or ** OldStartingIndex**. Shouldn't we be doing that across the board for consistency??

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
```diff
@@ -0,0 +1,96 @@
...
+								startIndex++;
+							}
+						} else {
+							PopulateComboBoxItems (items);
+						}
+					}
+					break;
+
+					case NotifyCollectionChangedAction.Remove:
+						if (e.OldItems != null) {
```
ermau: If there's no information about what items were removed, this code just acts like nothing happened, we need to treat it as a reset.

CartBlanche:
@ermau Latest push fixes the double/triple/quadruple enter bug, which I finally tracked down to being caused by calling `Cell.AccessibilityExpanded`. I'm now doing something more sensible and telling the selector to the `popUp:` if we aren't expanded.

CartBlanche:
@ermau Not sure if this has dropped off your radar or if you are awaiting for me to do something, that I'm not aware I should be working on with this PR.

ermau:
My last comment still hasn't been addressed in any way
https://github.com/xamarin/Xamarin.PropertyEditing/pull/394#discussion_r252424848

CartBlanche:
Whoa, totally missed that!

CartBlanche:
I've added a check in there to `Reset` if we have no info.

CartBlanche:
Index and null checks to `Reset()` now exist for both `Add` and `Remove` cases.

Xamarin.PropertyEditing.Mac/Controls/Custom/AutocompleteComboBox.cs
CartBlanche: Done.

CartBlanche:
@ermau Once the PopOver loses focus, it will jump back to the last focused control. Eg. If Filter has focus and you then pop open the CustomExpression, back out, it will jump back to the Filter text box. So are we saying that if you click a `PropertyButton`, it should first focus that row’s controls, before the PopOver appears, so that should you cancel/accept it it focuses on that row’s control?

PR #396, opened 2018-09-11
[Mac] Up-Down buttons will no longer allow interaction when disabled.

### LINKED ISSUE #395: [Mac] The Up/Down buttons still allow interaction, when they're supposed to be ReadOnly.

CartBlanche:
Fixes #395


Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableButton.cs
```diff
@@ -33,25 +33,31 @@ protected override void Dispose (bool disposing)
 
 		public override void MouseDown (NSEvent theEvent)
 		{
-			switch (theEvent.Type) {
+			if (Enabled) {
```
ermau: I'd recommend inverting stuff like this so instead of indenting everything, we simply have a
```csharp
if (!Enabled)
    return;
```
at the top

PR #412, opened 2018-09-24
[Mac] Use EditorViewModel in base locations

ermau:


ermau:
This is fixed fully in my grouped branch, so I'll just leave it to that instead of trying to pull it out.

Issue #390, opened 2018-09-05
[Mac] When selecting Reset for Point, Size and Rectangle controls, it resets per control.


CartBlanche:
It should reset all 4 values in 1 go. Windows works correctly.

Issue #400, opened 2018-09-13
Mac brush editor's current color and last color are identical


bleroy:
The current color should change while the cursor is being moved around in the shade and hue pickers, while the last color should remain constant, and should only switch to the new current when the user releases the mouse.

PR #414, opened 2018-09-25
Don't notify changes for any value change of a spin editor

### LINKED ISSUE #407: Switching between color formats is value unstable
Switching to HLS or HSB shifts the value still on Mac

bleroy:
Avoids premature commits to last color. Fixes #407 and #400.


CartBlanche:
Looks like this fixes https://github.com/xamarin/Xamarin.PropertyEditing/issues/390 too.

PR #415, opened 2018-09-26
[Tests] Ensure async void exceptions propagate correctly

alanmcgovern:
This should ensure that exceptions thrown inside async void
methods do get rethrown as part of the Teardown of a given test.

This removes the eventhandlers on AppDomain.UnhandledException
and also removes the second SyncContext implementation which
did not capture/propagate the exceptions.


alanmcgovern:
With the first change we are now correctly propagating exceptions through to NUnit. This reveals several latent test failures which were previously not being detected/reported:

vs-failures

The second patch prevents unhandled exceptions from tearing down the test runner process when executing from the built-in VSM test runner:
vsm-failures

With both patches in place we have a consistent 23 failures being reported on Windows and Mac.

alanmcgovern:
The third patches reduces the failures down to 4 :)

CartBlanche:
This 4th patch fixes the failing tests.

Xamarin.PropertyEditing.Tests/PanelViewModelTests.cs
```diff
@@ -58,7 +58,6 @@ public async Task InteruptedEditorRetrievalResolvesCorrectlyItemAdded ()
 			await Task.Yield ();
 
 			Assert.That (vm.Properties, Is.Empty);
-			this.context.ThrowPendingExceptions ();
```
CartBlanche: I'm assuming this is the correct change as the ThrowPendingExceptions should be happening in the TearDown and now here.

alanmcgovern: Exactly!

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -439,12 +439,13 @@ private async void UpdateAutocomplete (string value)
...
 					if (list == null) {
-						list = results.Result;
+						list = await results;
 						common = new HashSet<string> (list);
 					} else
-						common.IntersectWith (results.Result);
+						common.IntersectWith (await results);
 				} while (tasks.Count > 0 && !cancel.IsCancellationRequested);
 
+				cancel.ThrowIfCancellationRequested ();
```
ermau: This needs a test prove the new behavior, can you add one for @alanmcgovern, @CartBlanche ?

CartBlanche: Sure, I'll give it a try.

Xamarin.PropertyEditing/ViewModels/PropertyViewModel.cs
```diff
@@ -480,24 +479,32 @@ private async void UpdateAutocomplete (string value)
...
-				} while (tasks.Count > 0 && !cancel.IsCancellationRequested);
-
-				cancel.ThrowIfCancellationRequested ();
-				this.autocomplete.Reset (list.Where (common.Contains));
+				await ResetAutocompleteCollection (this.autocomplete, tasks, cancel);
 			} catch (OperationCanceledException) {
 			}
 		}
 
+		public async Task ResetAutocompleteCollection (ObservableCollectionEx<string> collection, List<Task<IReadOnlyList<string>>> tasks, CancellationToken cancel)
```
CartBlanche: @ermau Not sure if this is the sort of thing you had in mind, but extracted this out to make it easier to isolate this functionality.

ermau: Not sure why this needed to be extracted (you don't need it to test the behavior), let alone public. What external consumer would ever invoke this, especially with these arguments?

ermau: Making sure the exception is thrown is _not_ the thing to be testing here, it's the completion results when the results finished, but it was canceled.

CartBlanche: As your comment was literally on the newly added `cancel.ThrowIfCancellationRequested ();` I thought we wanted to ensure the exception was thrown, if the cancelation token's `IsCancelationRequested` was true. 

OK, So just to double check... We want to test that AutocompletionItems has *not* been updated when `GetCompletionsAsync` is cancelled??

If I'm still going down the wrong path, maybe a Slack call is in order?

ermau: > As your comment was literally on the newly added cancel.ThrowIfCancellationRequested (); I thought we wanted to ensure the exception was thrown, if the cancelation token's IsCancelationRequested was true.

There's an expected output, the contents of the autocomplete list. Exiting the method before the contents can be changed via the reset (by using `cancel.ThrowIfCancellationRequested()`) changes that expected output. 

> OK, So just to double check... We want to test that AutocompletionItems has not been updated when GetCompletionsAsync is cancelled??

Yes by way of the cancellation that's triggered if they start entering more text before the previous autocomplete requests can finish. Here's a demonstration in the change in behavior, in a case where you have 2 object editors.

1. First object editor's `GetCompletionAsync()` completes and returns items.
2. New text is entered, cancelling the second request and exiting the loop.
3. The list resets with the contents of the aggregate list, which will only contain items from 1.

With the throw before reset, #3 no longer occurs, living the auto-complete list to be whatever it was previously. That's the change in behavior to be tested for.

Xamarin.PropertyEditing.Tests/PropertyViewModelTests.cs
```diff
@@ -1175,6 +1175,64 @@ public void AutocompleteCancels ()
...
+				});
+
+			var mockProvider = new MockEditorProvider (editor1.Object);
+			var resources = new MockResourceProvider ();
+			var platform = new TargetPlatform (mockProvider, resources) {
+				SupportsCustomExpressions = true,
+			};
+
+			var vm = GetViewModel (platform, property.Object, new[] { editor1.Object, editor2.Object });
+			Assert.That (vm.SupportsAutocomplete, Is.True);
```
ermau: This should be an `Assume`

Xamarin.PropertyEditing.Tests/PropertyViewModelTests.cs
```diff
@@ -1175,6 +1175,64 @@ public void AutocompleteCancels ()
...
+
+			vm.PreviewCustomExpression = "preview";
+
+			changeResultSet = true;
```
ermau: We should add an `Assume` here to ensure that the results are the first set we expect them to be.

Xamarin.PropertyEditing.Tests/PropertyViewModelTests.cs
```diff
@@ -1175,6 +1175,64 @@ public void AutocompleteCancels ()
...
+
+			vm.PreviewCustomExpression = "preview2";
+
+			CollectionAssert.AreEqual (vm.AutocompleteItems, results);
+			complete1.Verify (c => c.GetCompletionsAsync (property.Object, "preview", It.IsAny<CancellationToken> ()));
```
ermau: Not sure this is relevant to this test

PR #416, opened 2018-09-28
Input modes

### LINKED ISSUE #382: Value input modes core
image

### LINKED ISSUE #383: Value input mode Windows
Windows implementation of #382 

Existing design:
image

ermau:
Closes #382, closes #383

### DIFF
--- a/Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
+++ b/Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
@@ -16,6 +16,7 @@ namespace Xamarin.PropertyEditing.Tests.MockControls
 			AddProperty<int> ("UnsetInteger", ReadWrite, valueSources: ValueSources.Local);
 			AddProperty<float> ("FloatingPoint", ReadWrite);
 			AddProperty<string> ("String", ReadWrite, valueSources: ValueSources.Local | ValueSources.Resource | ValueSources.Binding);
+			AddProperty<int> ("Width", ReadWrite, valueSources: ValueSources.Local | ValueSources.Resource | ValueSources.Binding, inputModes: new[] { new InputMode("Auto", true), new InputMode("Star"), new InputMode("Pixel"), });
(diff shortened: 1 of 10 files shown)


Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
```diff
@@ -16,6 +16,7 @@ public MockSampleControl ()
 			AddProperty<int> ("UnsetInteger", ReadWrite, valueSources: ValueSources.Local);
 			AddProperty<float> ("FloatingPoint", ReadWrite);
 			AddProperty<string> ("String", ReadWrite, valueSources: ValueSources.Local | ValueSources.Resource | ValueSources.Binding);
+			AddProperty<int> ("Width", ReadWrite, valueSources: ValueSources.Local | ValueSources.Resource | ValueSources.Binding, inputModes: new[] { new InputMode("Auto", true), new InputMode("Star"), new InputMode("Pixel"), });
```
bleroy: Either those are going to surface to the user, and they need to be localizable, or they're magic strings, and could probably be replaced with an enum.

Xamarin.PropertyEditing/ViewModels/NumericPropertyViewModel.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
bleroy: Oops?

Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
ermau: They're up to the provider to provide. In theory they will likely correspond to fixed names from code (enums, const properties) that should not be translated.

bleroy: Ah yes, this is just the mock implementation, indeed.

PR #419, opened 2018-10-04
Change numeric spin editor value changes on enter and tab, fix layout crashes

bleroy:
This fixes the problem where hitting enter and tabbing out of a numeric spin control (including numeric properties, size, point, rectangle) without breaking recent color picker fixes.

This also fixes layout issues that cause hard crashes in some mac configurations.


bleroy:
Oh wow, that is weird.

bleroy:
Layout changes moved to https://github.com/xamarin/Xamarin.PropertyEditing/tree/bleroy-mac-color-picker-layout-crashes

PR #440, opened 2018-10-17
Improve icon retrieval from brush tab view controller

bleroy:
The controller goes through the theme manager, and gracefully skips the icons if they're not found, instead of throwing from the depths of AppKit.


CartBlanche:
I didn't realise I'd missed this! Nicely spotted.

PR #441, opened 2018-10-23
[Mac] Set correct ValueSources for Brushes.

CartBlanche:
This also ensures that values entered for CustomExpression are stored and retrieved correctly.


CartBlanche:
@ermau I noticed that CustomExpression property on the Colour Editors doesn't save its values, because by default all the editors have `ValueSources.Default` set. Therefore this check...
https://github.com/xamarin/Xamarin.PropertyEditing/blob/81d11bc917abc250954a04f9121006d4096152f6/Xamarin.PropertyEditing.Tests/MockObjectEditor.cs#L179
always evaluate to true, so it never saves the entered CustomExpression value. Hence the question I posted on #proppy on Monday.
If `ValueSources.Default` should be included, then we'll need to change the check appropriately and I'll close this PR.

ermau:
It should be included, those aren't mutually exclusive features.

Issue #450, opened 2018-11-16
[Mac] Latest commits have broken control alignment.


CartBlanche:
As per screenshot...
<img width="659" alt="screen shot 2018-11-15 at 20 15 02" src="https://user-images.githubusercontent.com/271363/48615030-a4d3c980-e987-11e8-803b-d3b54f516034.png">

PR #451, opened 2018-11-16
[Mac] Fix Alignments so they are uniform across controls.

CartBlanche:
Fixes https://github.com/xamarin/Xamarin.PropertyEditing/issues/450

Before:
screen shot 2018-11-15 at 20 15 02

After:
screen shot 2018-11-16 at 15 36 07


CartBlanche:
@ermau If you don't have any objections I'll pull this in today?

CartBlanche:
@ermau I don't believe it is cut-off, that is just the way MacOS renders. 
Here is a screenshot from an app built in XCode.

screen shot 2018-11-20 at 21 03 13

I've updated the screenshot in the PR so it is easier to see the alignment. Also noticed some width alignment issues, which are now fixed.

One thing I found is that MacOS shifts controls in different directions by a few pixels, dependent on control, when you make them read-only. Why? IHNFI :/

CartBlanche:
@bleroy Please re-review.

PR #454, opened 2018-11-19
[Mac] Fixes 443. Only place where theming image isn't used initially.…

### LINKED ISSUE #443: Brush icons sometimes do not load
screen shot 2018-11-06 at 11 13 54 am

1. Select Mocked 1
1. Open ColorNoBrush's popup

CartBlanche:
… Handled later during selection.

Fixes #443


ermau:
I don't follow these changes, why would we move _away_ from the centralized system we put in place? If the images are reloaded elsewhere, why would we load them here at all?

CartBlanche:
@ermau The image loaded here is then used in the NSOutlineView's GetView call here
https://github.com/xamarin/Xamarin.PropertyEditing/blob/20ef532e71217711e83af4ec89c8fc69cddf9ce9/Xamarin.PropertyEditing.Mac/Controls/Custom/UnderlinedTabViewController.cs#L112

To then make sure we have the right image when the user selects or deselect the tabs. Done during the DrawRect here - 

https://github.com/xamarin/Xamarin.PropertyEditing/blob/a7b11a56e99a00177c705031b65de932f217dd11/Xamarin.PropertyEditing.Mac/Controls/Custom/UnderlinedTextField.cs#L55

As to why Larry went with that approach, you may need to ask him as that was the structure/flow in place when you approved his PR. Maybe we should look at refactoring it in future, but right now any teams who grab master have an unusable Brush view.

CartBlanche:
@bleroy To avoid the confusion with using the `NSImage` setting up front and avoid the unused image overhead, I've switched to storing the image name in the `NSTabViewItem.Identifier` object. Everything else appears to work as advertised. Please re-review.

Issue #433, opened 2018-10-11
Low-Frequency properties Core


ermau:
Properties that are hidden behind a fold as they are obscure.

PR #462, opened 2018-11-28
Support low-frequency properties

ermau:
This adds support for low-frequency properties in the core and Windows. Mac will arrange the properties to the bottom of a section but doesn't currently hide them.

xamarin propertyediting windows standalone_2018-11-16_14-30-58

Completes #433 
Completes #435

### DIFF
--- a/Xamarin.PropertyEditing/OrderedDictionary.cs
+++ b/Xamarin.PropertyEditing/OrderedDictionary.cs
@@ -30,11 +30,13 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+using Xamarin.PropertyEditing;
 
 namespace Cadenza.Collections
 {
 	internal class OrderedDictionary<TKey, TValue>
-		: IDictionary<TKey, TValue>, IList<KeyValuePair<TKey, TValue>>, IReadOnlyOrderedDictionary<TKey, TValue>
+		: IDictionary<TKey, TValue>, IList<KeyValuePair<TKey, TValue>>, IReadOnlyOrderedDictionary<TKey, TValue>, INotifyCollectionChanged
 	{
 		public OrderedDictionary ()
 			: this (0)
@@ -56,7 +58,7 @@ namespace Cadenza.Collections
 			this.dict = new Dictionary<TKey, TValue> (capacity, equalityComparer);
 			this.kvpCollection = this.dict;
 			this.keyOrder = new List<TKey> (capacity);
-			this.roKeys = new ReadOnlyCollection<TKey> (this.keyOrder);
+			this.roKeys = new ReadOnlyKeysCollection (this.keyOrder);
 			this.roValues = new ReadOnlyValueCollection (this);
 		}
 
@@ -74,6 +76,8 @@ namespace Cadenza.Collections
 			foreach (var kvp in dictionary)
 				Add (kvp.Key, kvp.Value);
 		}
+
+		public event NotifyCollectionChangedEventHandler CollectionChanged;
 
 		bool ICollection<KeyValuePair<TKey, TValue>>.IsReadOnly
 		{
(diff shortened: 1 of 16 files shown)


Xamarin.PropertyEditing/OrderedDictionary.cs
```diff
@@ -137,8 +155,17 @@ public IEqualityComparer<TKey> Comparer
 			get { return new KeyValuePair<TKey, TValue> (this.keyOrder[index], this[index]); }
 			set
 			{
-				keyOrder[index] = value.Key;
+				TKey existingKey = this.keyOrder[index];
+				if (!Equals (existingKey, value.Key))
+					this.keyOrder[index] = value.Key;
```
bleroy: This looks wrong. Let's say the dictionary has `{A, B}, {C, D}}` and you do `this[0] = new KVP {C, E}`, wouldn't you get `{{A, B}, {C, E}}` in the dictionary, `[C, C]` in the key order, and notifications that key `A` was replaced with `C` and that `{A, D}` was replaced with `{C, E}`?

Xamarin.PropertyEditing/OrderedDictionary.cs
```diff
@@ -303,6 +333,8 @@ public void Insert (int index, TKey key, TValue value)
 		{
 			this.keyOrder.Insert (index, key);
 			this.dict.Add (key, value);
```
bleroy: Not new to this change, but I think this can also throw if you try to insert a key that already exists.

Xamarin.PropertyEditing/OrderedDictionary.cs
```diff
@@ -333,7 +377,11 @@ public bool Remove (TKey key)
 		public void RemoveAt (int index)
 		{
 			TKey key = this.keyOrder[index];
```
bleroy: Should we document the possible out of range exception, like it's done on other methods?

Xamarin.PropertyEditing/OrderedDictionary.cs
```diff
@@ -333,7 +377,11 @@ public bool Remove (TKey key)
...
-			Remove (key);
+
+			this.keyOrder.RemoveAt (index);
+			this.dict.TryRemove (key, out var value);
```
bleroy: I'd defend in depth against the entry not being in the dictionary with an `Assert`.

Xamarin.PropertyEditing/OrderedDictionary.cs
ermau: It should, duplicate keys are invalid.

Xamarin.PropertyEditing/OrderedDictionary.cs
ermau: The try remove isn't defensive, it's to also grab the value for use in the `INCC` event.

Xamarin.PropertyEditing/OrderedDictionary.cs
bleroy: Yes, but in a way that potentially corrupts state if that exception is handled up the call stack. What I'm saying is that it may be better to explicitly handle the case up-front and throw before you attempt to change state. Otherwise, I think the `Insert` may succeed and the `Add` fail, resulting in inconsistent state.

ermau: Ah yeah fair point, I can swap their order.

Xamarin.PropertyEditing/OrderedDictionary.cs
bleroy: Sure, what I'm saying is that it should never happen, so I'd `Assert` on the result of the `TryRemove`. Also, there may be a possibility for inconsistent state here as well if that happened.

Xamarin.PropertyEditing/OrderedDictionary.cs
ermau: In fact, swapping them just produces the same concern in a different way (now key is added if index is out of range.) We have to check both prior.

PR #464, opened 2018-11-28
[Mac]Fix vertical alignment of labels in resource lists.

### LINKED ISSUE #463: [Mac] Text alignment in Resource Lists is no longer centred vertically.
Brushes
screen shot 2018-11-28 at 21 15 05

Resources
screen shot 2018-11-28 at 21 15 31

CartBlanche:
Fixes #463 

Brushes:
screen shot 2018-11-28 at 21 20 19

Resources:
screen shot 2018-11-28 at 21 20 48

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/ResourceOutlineViewDelegate.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/ResourceOutlineViewDelegate.cs
@@ -37,7 +37,6 @@ namespace Xamarin.PropertyEditing.Mac
 					var utf = (UnfocusableTextField)outlineView.MakeView (labelIdentifier, this);
 					if (utf == null) {
 						utf = new UnfocusableTextField {
-							BackgroundColor = NSColor.Clear,
 							Identifier = labelIdentifier,
 						};
 					}
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
@@ -33,6 +33,11 @@ namespace Xamarin.PropertyEditing.Mac
 			internal set { this.label.StringValue = value; }
 		}
 
+		public NSColor TextColor {
+			get { return this.label.TextColor; }
+			internal set { this.label.TextColor = value; }
+		}
+
 		public UnfocusableTextField ()
 		{
 			SetDefaultTextProperties ();
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
@@ -1,35 +1,100 @@
 ﻿using System;
-using System.Collections;
-using System.ComponentModel;
 using AppKit;
 using CoreGraphics;
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	class UnfocusableTextField : NSTextField
+	internal class UnfocusableTextField : NSView
 	{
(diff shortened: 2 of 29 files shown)


Xamarin.PropertyEditing.Mac/Controls/Custom/ResourceOutlineViewDelegate.cs
```diff
@@ -34,10 +34,9 @@ public override NSView GetView (NSOutlineView outlineView, NSTableColumn tableCo
 
 				case ResourceOutlineView.ResourceNameColId:
 				default:
-					var utf = (UnfocusableTextField)outlineView.MakeView (labelIdentifier, this);
+					var utf = (VerticallyCentredUnfocusableTextField)outlineView.MakeView (labelIdentifier, this);
```
bleroy: Nit: I think we usually standardize on US spelling, which would be `VerticallyCenteredUnfocusableTextField` here.

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
```diff
@@ -32,4 +32,30 @@ void SetDefaultProperties ()
 			BackgroundColor = NSColor.Clear;
 		}
 	}
+
+	internal class VerticallyCentredUnfocusableTextField : NSView
```
bleroy: It's not entirely clear why `VerticallyCentredUnfocusableTextField` doesn't derive from `UnfocusableTextField` as the name would imply, and as it seems to semantically be. Can we really not add those constraints otherwise?

bleroy: Also, without looking in too much details, I looked at `UnfocusableTextField` usage and it seems like we use it mostly as a label, and I'm not sure I saw a lot of cases where it doesn't need to be vertically centered. Even if we did, would it not be simpler to activate those constraints based on some `VerticallyCentered` boolean property on a single class, instead of having two separate controls that only differ by some minor attribute?

CartBlanche: We don't always want it to be vertically centred. For example in the case of the main PropertyListView, it essentially needs to be Top aligned. In the case of various controls (like Point, Rectangle etc) it isn't constrained because the view contains several controls so the Unfocusable version is used there and not this new one. We only need it to be vertically aligned when added to table rows, for now.

AFAICT it cannot be in 1 NSTextField or a subclass because NSTextField cannot centre itself, without custom drawing. Hence why it is nbow wrapped inside a NSView, where it is then centred vertically. I'm not happy with how that works, but Eric wasn't happy with my original solution of overriding the NSTextField DrawRect to position the text correctly, and he implemented this constraints solution. I'm just reusing his code, refactoring it, so that it can be used in more than 1 place.

bleroy: Fair enough, but my other suggestion for a way to manage that would be to have a unique control, using the `NSView` wrapping, but only doing the constraints if a `VerticallyCentered` property is true. If that's not practical or efficient, I'd still rename to clear the inheritance confusion, maybe to `VerticallyCenteredUnfocusableTextView`.

CartBlanche: Ah, Yes, that is possible. Not thinking straight.

CartBlanche:
@bleroy Everything now uses the 1 `UnfocusedTextView`.
There's now an extra commit to remove the `this.DoConstraints` call which has some performance overhead and these have been translated into direct `this.AddConstraints` call with `NSLayoutConstraint.Create(...)`. Which means proppy should initialise a little faster now.

PR #467, opened 2018-12-03
Improve keyboard navigation for brush editors

bleroy:
This does not fix keyboard navigation inside the popup, but makes it possible to open the popup using the keyboard once the brush property is focused.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BrushEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BrushEditorControl.cs
@@ -30,8 +30,20 @@ namespace Xamarin.PropertyEditing.Mac
 			if (Popover == null)
 				return;
 
-			if (!Popover.Shown)
+			if (!Popover.Shown) {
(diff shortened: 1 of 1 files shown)


CartBlanche:
@bleroy I'm not seeing the brush-popup receive control here. Maybe a small gif incase I'm doing something I shouldn't.

bleroy:
That's because nothing *inside* the pop-up is ready to receive focus. That's not what this change is doing. This enables the user to open the pop-up and close it using the keyboard. Keyboard navigation inside the pop-up will come later.

CartBlanche:
I wasn't talking about the navigation within the pop-up, but the control itself, in the property grid. I'm not seeing it receive focus.

bleroy:
Is your Mac configured to stop on all controls?

CartBlanche:
Good question. It does stop on all the others controls in the grid, just not the popup within it.

bleroy:
Before:


After:

PR #469, opened 2018-12-06
Workbooks: Prevent NRE on CALayer.ContentsFormat

sandyarmstrong:
It seems that in this branch, using `EnumPropertyViewModel` causes no
editor to be created. `PropertyTableDelegate` does not check for null
editors, resulting in NREs. Add checks, just in case.

In my testing, only `CALayer.ContentsFormat` ends up using
`EnumPropertyViewModel`. For now, comment-out the section in
`PropertiesViewModel.GetViewModel` that creates `EnumPropertyViewModel`,
so that at least a `StringPropertyViewModel` is created.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
@@ -87,6 +87,9 @@ namespace Xamarin.PropertyEditing.Mac
 					if (editor == null) {
 						editor = GetEditor (vm, outlineView);
 					}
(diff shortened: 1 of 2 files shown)


sandyarmstrong:
Please note that this is only meant for the (very old) `workbooks` branch. Ideally we'll bump to a recent release soon but I needed a quick fix for the crash.

sandyarmstrong:
Not planning to merge until I've tested on Windows (already tested on Mac).

sandyarmstrong:
Fine on Windows. Going to merge this.

PR #476, opened 2018-12-13
Adds implementation to handle TimeSpan types

netonjm:
This commit additionally creates a reusable ValidationTextField which needs some validation delegate to work
to handle internally.
The idea of this is reuse between the TextEditors who need a simple validation check.


Xamarin.PropertyEditing.Mac/Controls/Custom/ValidationTextField.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+	public interface IValidationTextDelegate
+	{
+		bool IsValid (NSText textObject);
+	}
+
+	public class ValidationTextField : NSTextField
+	{
+		string cachedValueString;
+		NSText CachedCurrentEditor { get; set; }
```
bleroy: If this is private, why is it a property and not a field?

Xamarin.PropertyEditing.Mac/Controls/TimeSpanEditorControl.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections;
+using AppKit;
+using Xamarin.PropertyEditing.Mac.Resources;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public class TimeSpanValidationTextDelegate : IValidationTextDelegate
+	{
+		public bool IsValid (NSText textObject) => TimeSpan.TryParse (textObject.Value, out var result);
```
bleroy: How is this handling localized input?

Xamarin.PropertyEditing.Mac/Controls/TimeSpanEditorControl.cs
```diff
@@ -0,0 +1,78 @@
...
+				BackgroundColor = NSColor.Clear,
+				ControlSize = NSControlSize.Small,
+				Font = NSFont.FromFontName (DefaultFontName, DefaultFontSize),
+				StringValue = string.Empty,
+				TranslatesAutoresizingMaskIntoConstraints = false,
+			};
+
+			// update the value on keypress
+			StringEditor.Changed += (sender, e) => {
+				if (TimeSpan.TryParse (StringEditor.StringValue, out TimeSpan result)) {
```
bleroy: Isn't that duplicating work with the time span validation above?

Xamarin.PropertyEditing.Mac/Controls/TimeSpanEditorControl.cs
netonjm: AFAIK if is not set uses the CultureInfo.CurrentCulture of the thread. We do the same in the old GTK Editor. https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.PropertyGrid.Editors/TimeSpanEditor.cs#L84-L87

Xamarin.PropertyEditing.Mac/Controls/TimeSpanEditorControl.cs
netonjm: Yeah.. I'm too defensive

Xamarin.PropertyEditing.Mac/Controls/TimeSpanEditorControl.cs
bleroy: Can you check whatever we're doing for floating point types, so we're consistent?

netonjm: Of course sir!

PR #479, opened 2018-12-14
[Mac] Adds implementation of FilePathEditorControl

netonjm:
This PR adds a new Editor to the current Property Panel. It handles a new Type: FilePath which defines a Source property to define the path of the selected file.

The commit includes fixes proposed from the design team and an example in the  Standalone Designer with the new property.

Fixes VSTS #732636 - Implement FilePath editor

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/732636

openfolder2-1


CartBlanche:
Same as for the FilePath PR....
Again this might be a question for @vancura and the design time, but I'd be more in favour of this being an edit box for a a better UX, as I think devs sometimes like to type paths, or copy and paste a path into there. This might mean having a separate button on the right to click, like the XCode Font control (see screenshot below), which then pops up the DirectoryPath dialog.

screen shot 2018-12-14 at 11 24 12

Forcing devs to click, without alternatives, isn't a great UX IMO. Just a suggestion, but Vaclav and @ermau have final say.

vancura:
Yes, please make this a text field, as I want to paste paths from clipboard (or copy to clipboard). It needs to be selectable and editable.

vancura:
Let me draw how the button could work. Give me a minute.

vancura:
labels inputs mac 2x

See last two editors.

netonjm:
Ready to review! 👍

Xamarin.PropertyEditing.Mac/Controls/DirectoryPathEditorControl.cs
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		protected override void SetEnabled ()
+		{
+			ButtonEditor.Enabled = ViewModel.Property.CanWrite;
+		}
+
+		protected override void UpdateAccessibilityValues ()
+		{
+			ButtonEditor.AccessibilityTitle = string.Format (LocalizationResources.AccessibilityString, ViewModel.Property.Name);
```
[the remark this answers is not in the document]
netonjm: Fixed

Xamarin.PropertyEditing.Mac/Controls/Custom/TextFieldSmallButtonContainer.cs
```diff
@@ -0,0 +1,139 @@
...
+			TranslatesAutoresizingMaskIntoConstraints = false;
+			Cell.HighlightsBy = (int)NSCellStyleMask.ContentsCell;
+		}
+	}
+
+	internal class TextFieldSmallButtonContainer : NSTextField
+	{
+		public List<SmallButton> Buttons { get; } = new List<SmallButton> ();
+		private readonly ButtonTextFieldCell cell;
+		internal NSView lastView;
```
ermau: This doesn't seem to need to be exposed

Xamarin.PropertyEditing/Common/FilePath.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace Xamarin.PropertyEditing
+{
+	public class FilePath
+	{
+		public string Source { get; set; }
+		public bool IsDirectory { get; set; }
```
ermau: You can't actually depend on something from the value to determine the behavior of the control as you may not have a value to begin with. File and Directory path need to be separate types to distinguish behaviors. The base of the controls can largely share behavior with a couple overrides to define file or directory specifics.

Xamarin.PropertyEditing/Common/FilePath.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace Xamarin.PropertyEditing
```
ermau: This should have the `.Common` namespace.

Xamarin.PropertyEditing/Common/FilePath.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+	public class FilePath
```
ermau: This should implement equality.

Xamarin.PropertyEditing/Common/FilePath.cs
netonjm: done.

Xamarin.PropertyEditing/Common/FilePath.cs
netonjm: done

Xamarin.PropertyEditing/Common/FilePath.cs
netonjm: done

Xamarin.PropertyEditing.Mac/Controls/Custom/TextFieldSmallButtonContainer.cs
netonjm: done

netonjm:
@ermau  could you please take a look at the changes? I dropped all the code related to the directory. Please check if all is ok for you and once merged I will create a new PR with directory implementation.

Xamarin.PropertyEditing.Mac/Controls/Custom/TextFieldSmallButtonContainer.cs
```diff
@@ -0,0 +1,139 @@
...
+			ControlSize = NSControlSize.Small;
+			Bordered = false;
+			TranslatesAutoresizingMaskIntoConstraints = false;
+			Cell.HighlightsBy = (int)NSCellStyleMask.ContentsCell;
+		}
+	}
+
+	internal class TextFieldSmallButtonContainer : NSTextField
+	{
+		public List<SmallButton> Buttons { get; } = new List<SmallButton> ();
```
ermau: There doesn't seem to be any reason for this to be public?

Xamarin.PropertyEditing.Mac/Controls/FilePathEditorControl.cs
```diff
@@ -0,0 +1,166 @@
...
+using CoreGraphics;
+using Foundation;
+using Xamarin.PropertyEditing.Mac.Resources;
+using Xamarin.PropertyEditing.ViewModels;
+using System.IO;
+using Xamarin.PropertyEditing.Common;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class FilePathEditorControl : PropertyEditorControl<PropertyViewModel<FilePath>>
```
ermau: Most of this editor can be shared with a directory control, can we split it into a `abstract BasePathEditorControl` and a file implementation?

Xamarin.PropertyEditing/Common/FilePath.cs
```diff
@@ -0,0 +1,42 @@
...
+		public override bool Equals (object obj)
+		{
+			if (obj == null) return false;
+			if (!(obj is FilePath)) return false;
+			return Equals ((FilePath)obj);
+		}
+
+		public bool Equals (FilePath other)
+		{
+			return Source.Equals (other.Source, StringComparison.InvariantCultureIgnoreCase);
```
ermau: This will NRE if `other` is `null`

Xamarin.PropertyEditing/Common/FilePath.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Xamarin.PropertyEditing.Common
+{
+	public class FilePath : IEquatable<FilePath>
+	{
+		public string Source { get; set; }
+
+		public FilePath () { }
+
+		public FilePath (string source)
```
ermau: Let's add a `ArgumentNullException` here for `source`, otherwise we might crash later in `GetHashCode()` if it's null.

Xamarin.PropertyEditing/Common/FilePath.cs
netonjm: Fixed

Xamarin.PropertyEditing.Mac/Controls/FilePathEditorControl.cs
netonjm: done!

Xamarin.PropertyEditing.Mac/Controls/Custom/TextFieldSmallButtonContainer.cs
netonjm: Fixed

Xamarin.PropertyEditing/Common/FilePath.cs
netonjm: Fixed

netonjm:
@ermau I did the changes, could you please take a look? Thanks!

netonjm:
Branch rebased!

Xamarin.PropertyEditing.Mac/Controls/FilePathEditorControl.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+	internal class FilePathEditorControl : BasePathEditorControl<FilePath>
+	{
+		public FilePathEditorControl (IHostResourceProvider hostResource)
+		: base (hostResource)
+		{
+			this.currentTextField.ToolTip = this.currentTextField.PlaceholderString = string.Format (LocalizationResources.ChooseFileOrDirectory, LocalizationResources.File);
+			this.panel.CanChooseFiles = true;
+			this.panel.CanChooseDirectories = false;
+			this.revealPathButton.ToolTip = string.Format (LocalizationResources.RevealFileOrDirectory, LocalizationResources.File);
```
ermau: Does Voice Over read these if there's no accessibility titles? Perhaps we should have titles.

Xamarin.PropertyEditing.Mac/Resources/LocalizationResources.resx
```diff
@@ -122,4 +122,28 @@
     <value>{0} InputMode Editor</value>
     <comment>Editor for InputMode Value</comment>
   </data>
+  <data name="AccessibilityPathEditor" xml:space="preserve">
+    <value>{0} Path Editor</value>
```
ermau: This is missing the actual property name itself.

Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
```diff
@@ -0,0 +1,121 @@
...
+			this.revealPathButton.Activated += OnRevealPathButtonActivated;
+
+			#endregion
+
+			#region Browse Path
+
+			this.browsePathButton = new SmallButton {
+				Font = NSFont.FromFontName (DefaultFontName, DefaultFontSize),
+				StringValue = string.Empty,
+				Image = hostResource.GetNamedImage ("path-browse"),
```
ermau: These images need to be set in `ViewDidChangeEffectiveAppearance` so they can properly adapt for dark/light on change and the method should be called at the bottom of the `ctor`

Xamarin.PropertyEditing.Mac/Controls/FilePathEditorControl.cs
netonjm: In NSButtons the accessibility.title is by default the title of the button AFAIK

Xamarin.PropertyEditing.Mac/Resources/LocalizationResources.resx
netonjm: Fixed

Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
netonjm: Fixed

Xamarin.PropertyEditing.Mac/Controls/FilePathEditorControl.cs
netonjm: Yep. But we can include it if you want.

filepath

CartBlanche:
I’ve spotted some usability issues(clipping, enabling, reseting) and @netonjm and I are working on them.

netonjm:
@ermau the PR is ready for revision

Xamarin.PropertyEditing/Common/FilePath.cs
ermau: You fixed this particular request, but it can still NRE in `GetHashCode()` since `Source` is settable and we have an argumentless `ctor`. Since `FilePath` itself is a reference type, I don't think we need to support a null `Source`, so let's just make `FilePath` immutable.

Xamarin.PropertyEditing.Mac/Controls/FilePathEditorControl.cs
```diff
@@ -0,0 +1,60 @@
...
+		protected override void OnRevealPathButtonActivated (object sender, EventArgs e)
+		{
+			if (File.Exists (this.currentTextField.StringValue)) {
+				NSWorkspace.SharedWorkspace.OpenFile (this.currentTextField.StringValue);
+			}
+		}
+
+		protected override void StoreCurrentValue ()
+		{
+			if (ViewModel.Value == null) {
```
ermau: Why would we back out if the existing value is null? This just means we can never set a value to an unset value of this property. If you comment out the initial value setter you have you'll see nothing is set when you try to set it.

CartBlanche: This was one of the things I fixed in my PR and is what caused the enablement issue.

ermau:
> After I set a value, the browse to button never becomes enabled.
This hasn't been addressed.

Additionally, you're going to want to make the sizing and layout changes appropriate to #546 as it will likely land before this does. Guidelines can be found at the bottom of the commit message here https://github.com/xamarin/Xamarin.PropertyEditing/commit/ca63f0d6edb622113c3e09bb4590b42edef73e0c

vancura:
Could you please show me a full-res lossless screenshot so I can measure? Also dark theme, please.

CartBlanche:
@ermau That was definitely fixed in a previous PR. Might be a regression. I'll double check.

ermau:
> @ermau That was definitely fixed in a previous PR. Might be a regression. I'll double check.

I've checked it again, it doesn't work.

CartBlanche:
Quick video of this working for me. Unless we are doing different things. I'm on commit d775f95d0ecf259ec402792770297eb58416d867

screen recording 2019-03-06 at 14 33 52

CartBlanche:
Quick video of this working for me. Unless we are doing different things. I'm on commit ca727f19e85293454530e6e2c4e9e5d2158e5113 after the recent rebase.

Video accessible from this link...
https://1drv.ms/v/s!Agbp4B3CE3TVgpxaimEn9IqgAjxzrQ

ermau:
So what's happening in my case is that I've been select an `.xcodeproj` file, which is actually a package. macOS' file selector see it as a file so you can pick it, but the .NET file APIs see it as a directory.

Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
```diff
@@ -0,0 +1,148 @@
...
+			this.currentTextField.AccessibilityRole = NSAccessibilityRoles.TextFieldRole;
+			this.revealPathButton.AccessibilityRole = NSAccessibilityRoles.ButtonRole;
+			this.browsePathButton.AccessibilityRole = NSAccessibilityRoles.ButtonRole;
+
+			#endregion
+
+			AddConstraints (new[] {
+				NSLayoutConstraint.Create (this.currentTextField, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this, NSLayoutAttribute.Left, 1, 0f),
+				NSLayoutConstraint.Create (this.currentTextField, NSLayoutAttribute.CenterY, NSLayoutRelation.Equal, this, NSLayoutAttribute.CenterY, 1, 0f),
+				NSLayoutConstraint.Create (this.currentTextField, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this, NSLayoutAttribute.Width, 1, -33f),
```
ermau:
```suggestion
				NSLayoutConstraint.Create (this.currentTextField, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this, NSLayoutAttribute.Width, 1, 0f),
```
As of #546  `EditorContainer` now handles the property button, so we don't need to size for it.

Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
```diff
@@ -0,0 +1,148 @@
...
+				NSLayoutConstraint.Create (this.currentTextField, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1, DefaultControlHeight),
```
ermau:
```suggestion
				NSLayoutConstraint.Create (this.currentTextField, NSLayoutAttribute.Height, NSLayoutRelation.Equal, this, NSLayoutAttribute.Height, 1, -6),
```

netonjm:
> .xcodeproj

good catch 
SetTreatsFilePackagesAsDirectories:YES

CartBlanche:
@ermau Rebased and Suggestions applied.

ermau:
Now it seems that the navigate to button is opening the file, rather than opening finder to that file.

CartBlanche:
@ermau I've changed the original code to OpenFinder and Select the File.
I'd also like to propose that we change the tooltip text from "Reveal File" to "Show in Finder" as that is the standard vocabulary used throughout MacOS, AFAICK.

alanmcgovern:
@sevoku

PR #480, opened 2018-12-14
[Mac] Adds implementation to handle DateTime fields

netonjm:
This PR makes possible handle DateTime fields in Proppy. 

I changed to use a NSDatePicker instead a simple TextField with validation.

datetime


Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
```diff
@@ -0,0 +1,71 @@
...
+		{
+			editor = new DateTimeTextField {
+				Font = NSFont.FromFontName (DefaultFontName, DefaultFontSize),
+			};
+
+			// update the value on keypress
+			editor.Changed += (sender, e) => {
+				if (DateTime.TryParse (editor.StringValue, out DateTime dateValue)) {
+					ViewModel.Value = dateValue;
+					UpdateValue ();
```
CartBlanche: you shouldn't need to call `UpdateValue` here. Assigning `ViewModel.Value` should automagically call `UpdateValue` under the hood.

CartBlanche:
Just for consistency, suggesting the possible use of the built in MacOS DateTime control which has the built in spinner. But again subject to approval from @vancura and @ermau.

Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
netonjm: mm... then this hiddes an issue updating viewmodel, because I added because was not refreshing it.

CartBlanche: If you look at the `StringEditorControl`, or any of the other editors, there are no calls to `UpdateValue` after `ViewModel.Value` is set. So I agree, there must be something else going on that is breaking the correct set of events.

netonjm:
@vancura thoughts? ❤️

vancura:
Looks good on the GIF, but could you please grab a fullres @2x PNG? I am afraid there's some clipping around.

Xamarin.PropertyEditing.Mac/Controls/DateExtensions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using Foundation;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public static class DateExtensions
```
ermau: This should be internal.

Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
```diff
@@ -0,0 +1,83 @@
...
+		{
+			this.editor.Enabled = ViewModel.Property.CanWrite;
+		}
+
+		protected override void UpdateAccessibilityValues ()
+		{
+			this.editor.AccessibilityTitle = string.Format (LocalizationResources.AccessibilityDateTime, ViewModel.Property.Name);
+		}
+
+		protected override void Dispose (bool disposing)
```
ermau: Is this actually necessary?

netonjm: This was a long discussion related to memory allocation leaks in Xamarin.Mac over handlers. I prefer to invoke @Therzok which know better all the bit and surely can explain in detail the reason for this. But yes, we have a rule in MD to remove all the handling in Dispose.

Xamarin.PropertyEditing.Mac/Controls/DateExtensions.cs
netonjm: 👍

Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
Therzok: @ermau there's scenarios where ref-cycles do happen and only an event handler removal does fix it.

Even though Xamarin.Mac has weak event handling, Gtk# does not, so if we get through a toggleref to toggleref cycle between Gtk# widgets via this, we will end up leaking.

Just adds safety and the possibility to fix leaks via explicit disposal.

Therzok: Afaik, the eventhandler is stored via strong references, the target is stored via weak ones. So we might still encounter a refcycle via eventhandlers even in pure Xamarin.Mac.

I think there's an okay explanation here: https://github.com/mono/monodevelop/wiki/UI-code---Reference-counting-and-you#xamarinmac

tl;dr: Best of any world is actually having a static selector/eventhandler exported and using that, to avoid any kind of captures that would lead to refcycles.

netonjm:
image

ermau:
Now there's a gap at the top of the control.

netonjm:
@ermau  I did some changes to the current implementation because DatePicker is not friendly to reduce its size. Then I needed to hack a little the control to match the design specs. Could you please take a look?
datetime

vancura:
Could you give me a static (not animated lossless) version of the image above? I'd like to measure it. Thanks!

netonjm:
image

Xamarin.PropertyEditing.Mac/Controls/DateExtensions.cs
```diff
@@ -32,8 +32,12 @@ public static NSDate ToNSDate (this DateTime dt)
 		/// <returns>A DateTime</returns>
 		public static DateTime ToDateTime (this NSDate nsDate)
 		{
-			// We loose granularity below millisecond range but that is probably ok
-			return _nsRef.AddSeconds (nsDate.SecondsSinceReferenceDate);
+			try {
+				// We loose granularity below millisecond range but that is probably ok
+				return _nsRef.AddSeconds (nsDate.SecondsSinceReferenceDate);
+			} catch (Exception) {
+				return _nsRef;
```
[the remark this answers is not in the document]
netonjm: Because in lower dates than minimum date it throws an Exception (I checked this)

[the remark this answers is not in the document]
ermau: What's the "minimum" in this case? Can we catch the _specific_ exception it throws instead of a blanket catch?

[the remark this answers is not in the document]
CartBlanche: @netonjm can we not check that the entered date isn't below minimum (and assume we need a similar check for above maximum) to avoid the exception and just return the min/max with a ding sound?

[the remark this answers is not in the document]
CartBlanche: @netonjm from some quick testing this morning it looks like an `ArgumentOutOfRangeException` is thrown, which makes sense as I tried to enter a negative value and sounds like it would be the most likely exception to be thrown in a Datetime field.
See screenshot below..
Screen Shot 2019-03-12 at 11 14 52
Is there another scenario where it is some other exception? If so, what steps are required to trigger this different scenario?

alanmcgovern:
@sevoku

PR #493, opened 2019-01-11
Revamp Mac theming

ermau:
This introduces new APIs for providing Mac theme images, colors and themes. Additionally it updates the brush editors to more closely (but not yet completely) match the design docs using these APIs (and some layout tweaks). **Checkout and test requested.**

Before:

screen shot 2019-01-11 at 12 17 29 pm

screen shot 2019-01-11 at 12 17 17 pm

After:

screen shot 2019-01-11 at 3 05 09 pm

screen shot 2019-01-11 at 12 13 40 pm

Dark mode will not look correct in the standalone in <=High Sierra. This is known.


Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
```diff
@@ -97,8 +111,16 @@ public void Select (IEnumerable<object> selectedItems)
 			((ObservableCollectionEx<object>)SelectedItems).Reset (selectedItems);
 		}
 
+		public override void ViewDidChangeEffectiveAppearance ()
+		{
+			this.propertyArrangeMode.Font = HostResourceProvider.GetNamedFont (PropertyEditorControl.DefaultFontName, PropertyEditorControl.DefaultFontSize);
```
CartBlanche: Ran the project for the first time and `this.property*` all all null, so getting an NRE exception on start-up.

bleroy:
On Mojave dark mode with the light theme selected, the color popup background isn't set and appears dark, making the selected tab icon hard to read.

bleroy:
Selecting "Resource" in the property dropdown for the material design brush in the standalone causes a missins method exception in `ResourceTableDelegate.SetUpRenderer` because it can't find the ctor for `CommonBrushView`.

bleroy:
Selecting "Resource" in the property dropdown for the material design brush in the standalone causes a missins method exception in `ResourceTableDelegate.SetUpRenderer` because it can't find the ctor for `CommonBrushView`.

bleroy:
Starting from a brush set to no brush, and selecting MD, we get multiple selections, the brush isn't getting updated: http://g.recordit.co/NpgtNN9vox.gif

bleroy:
Should up/down controls have a rollover effect when they are disabled? (check it out on read-only properties).

CartBlanche:
@bleroy I don't think they should, as they aren't interactive. I don't believe this isn't part of the @ermau's changes in this PR, but a pre-existing bug.

ermau:
There's still some slight differences in the design, but they're not worth blocking on. Former comments should be addressed. The brush issues exist in master.

PR #502, opened 2019-01-21
[Mac] Creates a new focusable ComboBox with readonly modifier

### LINKED ISSUE #497: [Mac][Accessibility] NSCombox doesn't allow focus with keyboard
When we are navigating with tab in the controls, some items (NSComboBox) are not in the Focus chain.

Any control should be focusable with the keyboard.

This happens with the properties and with the ArrangleBy combo.

focusissue

netonjm:
When we set NSComboBox.Editable to false, we lose all the signals from the NSView included focus.

I created a FocusableComboBox which doesn't allows modify the current text in the entry.

Fixes #497


CartBlanche:
Eric mentioned we may not need this, but I'm unsure when he had in mind this would no longer be needed.

Having it named as `Focusable*` suggests it might be editable. My only suggestion would be to rename it to `FocusableReadOnlyCombobox`, so the functionality is clearer.

ermau:
All I said was that the arrange by box was going away, not that combo boxes in general were.

netonjm:
@CartBlanche done!

netonjm:
@ermau done

Issue #488, opened 2018-12-15
[Mac] Images are missing in MonoDevelop (resource missing?)


netonjm:
![image](https://user-images.githubusercontent.com/1587480/50037813-9247c100-0015-11e9-9198-29495b3238fd.png)

It needs some investigation but looks like the resources are not in the expected folder for some reason.

CartBlanche:
After a chat with @netonjm This occurs because our resources are set to Bundle Resources. We'll need to change this to EmbeddedResource, unless some has another suggestion.

alanmcgovern:
Yes - we need to embed them as we don't have an appbundle.

If we use any NSBundle based APIs to load images we will have to migrate them too :)

ermau:
> After a chat with @netonjm This occurs because our resources are set to Bundle Resources. We'll need to change this to EmbeddedResource, unless someone has another suggestion.

This may break named NSImages, we'll need to update the default `HostResourceProvider` implementation to address the change.

CartBlanche:
@netonjm Can I close this issue? I believe you tested and merged my PR on the VS4Mac side.

netonjm:
Yes! now is a MD task

PR #516, opened 2019-02-01
Add pe- prefix to all image resources.

CartBlanche:
Partial Fix to https://github.com/xamarin/Xamarin.PropertyEditing/issues/488

MSBuild command has also been added to @netonjm's branch(https://github.com/mono/monodevelop/pull/6742), but this PR would need to go in first and then Jose will need to bump the submodule.


Xamarin.PropertyEditing.Mac/HostResourceProvider.cs
```diff
@@ -13,6 +13,13 @@ public NSAppearance CurrentAppearance
 			set;
 		}
 
+		private readonly NSBundle resourceBundle;
+
+		public HostResourceProvider ()
+		{
+			this.resourceBundle = new NSBundle ("../../../PropertyEditingResource.bundle");
```
alanmcgovern: We should change this to something like:
```
var containingDir = Path.GetDirectoryName (typeof (HostResourceProvider).Assembly.Location);
var bundle = Path.Combine (containingDir, "PropertyEditingResource.bundle");
```

Xamarin.PropertyEditing.Mac.Standalone/Xamarin.PropertyEditing.Mac.Standalone.csproj
```diff
@@ -114,4 +114,19 @@
...
+    <ItemGroup>
+      <PropertyEditingResourceBundlePath Include="$(SolutionDir)\Xamarin.PropertyEditing.Mac\PropertyEditingResource\**\*" />
+    </ItemGroup>
+    <Copy 
+      SourceFiles="@(PropertyEditingResourceBundlePath)" 
+      DestinationFiles="@(PropertyEditingResourceBundlePath->'$(OutDir)\PropertyEditingResource.bundle\%(RecursiveDir)%(Filename)%(Extension)')" 
+      SkipUnchangedFiles="true" 
+      OverwriteReadOnlyFiles="true" 
+      Retries="3" 
+      RetryDelayMilliseconds="300" />
```
alanmcgovern: If we made this part of `Xamarin.PropertyEditor.Mac.csproj` then it may be easier to ensure the files are copied everywhere they should be :)

Xamarin.PropertyEditing.Mac/Xamarin.PropertyEditing.Mac.csproj
```diff
@@ -170,162 +171,162 @@
...
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up-focus-blue~dark.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up-focus-blue~dark%402x.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up-focus-graphite.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up-focus-graphite%402x.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up-focus-graphite~dark.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up-focus-graphite~dark%402x.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up%402x.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up~dark.png" />
+    <None Include="PropertyEditingResource\Contents\Resources\pe-stepper-up~dark%402x.png" />
```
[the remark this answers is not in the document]
CartBlanche: I left in it so that from within the IDE you can see what resources you have, otherwise you have to keep flicking to the Resource Dir to see what is there. Also makes it easy to keep things in sync with git from within the IDE. ie, if you add a resource via the IDE, it will automagically add it to the git repo.

Xamarin.PropertyEditing.Mac.Standalone/Xamarin.PropertyEditing.Mac.Standalone.csproj
alanmcgovern: I tweaked it to replicate what we did in other places in the code, and the end result is this:

https://gist.github.com/alanmcgovern/d86dc42e1314db1f32f04f7cc3306caf

I think that ensures the files are always in the right place

Xamarin.PropertyEditing.Mac/HostResourceProvider.cs
CartBlanche: @alanmcgovern `HostResourceProvider`'s assembly ends up in `*.app/Contents/MonoBundle`, so we'd still need to `../../../` to get to the `Resource.bundle`. Should the copy put in the files in `*.app/Contents/MonoBundle` too, or not?

alanmcgovern: Yup! We probably still need a `cp` in standalone designer to copy the bundle from the output directory inside the app bundle itself. I missed that in my original gist!

CartBlanche:
I believe @netonjm has cherry picked this over to his MD branch and also merged the `DesignerSupport` changes too and confirmed it does what it says on the tin.

Xamarin.PropertyEditing.Mac.Standalone/Xamarin.PropertyEditing.Mac.Standalone.csproj
```diff
@@ -123,7 +123,7 @@
     </ItemGroup>
     <Copy 
       SourceFiles="@(PropertyEditingResourceBundlePath)" 
-      DestinationFiles="@(PropertyEditingResourceBundlePath->'$(AppBundleDir)\Contents\MonoBundle\PropertyEditingResource.bundle\%(RecursiveDir)%(Filename)%(Extension)')"
+      DestinationFiles="@(PropertyEditingResourceBundlePath->'$(OutputPath)\$(AssemblyName).app\Contents\MonoBundle\PropertyEditingResource.bundle\%(RecursiveDir)%(Filename)%(Extension)')"
```
CartBlanche: Thanks for the quick fix @alanmcgovern.

PR #520, opened 2019-02-05
Mac object editor

### LINKED ISSUE #141: Mac Object Editor
Depends on #116

 - [x] Core
 - Editor:
   - [x] Design
   - [x] Localized
   - Behavior:
     - [x] Keyboard control
     - [x] Accessibility
     - [x] Selection
   - Theming:
     - [ ] Light themed
     - [ ] Dark themed
     - [ ] High contrast
   - States:
     - [ ] Error state
     - [ ] Bound state
     - [ ] Conflict state
     - [ ] Pending state

ermau:
This PR:
 - Adds the object editor control for Mac (closes #141)
 - Reimplements the disclosure handling for the outline view allowing us to properly control backgrounds
 - Adds a reusable type selector which is needed for the both this editor and the binding editor
 - Fixes reuse for first instance of each editor type (first-sized)

screen shot 2019-02-05 at 3 45 40 pm

screen shot 2019-02-05 at 3 45 54 pm

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
@@ -18,14 +18,15 @@ namespace Xamarin.PropertyEditing.Mac
 		{
 			get { return viewModel; }
 			set {
-				if (viewModel != null) {
-					viewModel.PropertyChanged -= OnPropertyChanged;
+				if (this.viewModel != null) {
+					this.viewModel.PropertyChanged -= OnPropertyChanged;
 				}
 
-				viewModel = value;
-				viewModel.PropertyChanged += OnPropertyChanged;
-
-				ValueSourceChanged (viewModel.ValueSource);
+				this.viewModel = value;
+				if (this.viewModel != null) {
+					this.viewModel.PropertyChanged += OnPropertyChanged;
+					ValueSourceChanged (this.viewModel.ValueSource);
+				}
 			}
 		}
 
--- a/Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
@@ -43,6 +43,14 @@ namespace Xamarin.PropertyEditing.Mac
 			set { this.label.StringValue = value; }
 		}
 
+		public override void ViewWillMoveToSuperview (NSView newSuperview)
+		{
+			if (newSuperview == null && EditorView != null)
+				EditorView.ViewModel = null;
+
+			base.ViewWillMoveToSuperview (newSuperview);
+		}
+
 		private UnfocusableTextField label = new UnfocusableTextField {
 			Alignment = NSTextAlignment.Right,
 			TranslatesAutoresizingMaskIntoConstraints = false
@@ -53,7 +61,5 @@ namespace Xamarin.PropertyEditing.Mac
 			set { this.label.TextColor = value; }
 		}
 #endif
-
-		private readonly IHostResourceProvider hostResources;
 	}
 }
--- a/Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
@@ -43,6 +43,34 @@ namespace Xamarin.PropertyEditing.Mac
 			set { this.label.StringValue = value; }
 		}
 
+		public NSView LeftEdgeView
+		{
+			get { return this.leftEdgeView; }
+			set
+			{
+				if (this.leftEdgeView != null) {
+					this.leftEdgeView.RemoveFromSuperview ();
+					RemoveConstraints (new[] { this.leftEdgeLeftConstraint, this.leftEdgeVCenterConstraint });
+					this.leftEdgeLeftConstraint.Dispose ();
+					this.leftEdgeLeftConstraint = null;
+					this.leftEdgeVCenterConstraint.Dispose ();
+					this.leftEdgeVCenterConstraint = null;
+				}
+
+				this.leftEdgeView = value;
+
+				if (value != null) {
+					AddSubview (value);
+
+					value.TranslatesAutoresizingMaskIntoConstraints = false;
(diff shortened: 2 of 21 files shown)


Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -18,14 +18,16 @@ internal PropertyViewModel ViewModel
...
 				}
 
-				viewModel = value;
-				viewModel.PropertyChanged += OnPropertyChanged;
+				this.viewModel = value;
+				if (this.viewModel != null)
+					this.viewModel.PropertyChanged += OnPropertyChanged;
 
-				ValueSourceChanged (viewModel.ValueSource);
+				if (value != null)
```
bleroy: This is the exact same condition as above. Merge them?

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -43,6 +43,42 @@ public EditorContainer (IHostResourceProvider hostResources, IEditorView editorV
 			set { this.label.StringValue = value; }
 		}
 
+		public NSView LeftEdgeView
+		{
+			get { return this.leftEdgeView; }
+			set
+			{
```
bleroy: Lovely.

Xamarin.PropertyEditing.Mac/Controls/TypeSelectorControl.cs
```diff
@@ -0,0 +1,216 @@
...
+				NSLayoutConstraint.Create (scroll, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this.filter, NSLayoutAttribute.Width, 1, 0),
+				NSLayoutConstraint.Create (scroll, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, this, NSLayoutAttribute.CenterX, 1, 0),
+				NSLayoutConstraint.Create (scroll, NSLayoutAttribute.Bottom, NSLayoutRelation.Equal, this.checkbox, NSLayoutAttribute.Top, 1, -8),
+
+				NSLayoutConstraint.Create (this.checkbox, NSLayoutAttribute.Bottom, NSLayoutRelation.Equal, this, NSLayoutAttribute.Bottom, 1, -10),
+				NSLayoutConstraint.Create (this.checkbox, NSLayoutAttribute.Left, NSLayoutRelation.Equal, scroll, NSLayoutAttribute.Left, 1, 0)
+			});
+		}
+
+		public override void KeyDown (NSEvent theEvent)
```
bleroy: What's this doing?

Xamarin.PropertyEditing.Mac/Controls/TypeSelectorControl.cs
```diff
@@ -0,0 +1,216 @@
...
+			this.outlineView.DataSource = new TypeSelectorDataSource (ViewModel);
+			OnPropertyChanged (ViewModel, new PropertyChangedEventArgs (null));
+		}
+
+		public override void OnPropertyChanged (object sender, PropertyChangedEventArgs e)
+		{
+			base.OnPropertyChanged (sender, e);
+
+			switch (e.PropertyName) {
+			case nameof (TypeSelectorViewModel.FilterText):
```
bleroy: Since this is the same as default, can you remove it?

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -197,6 +249,25 @@ public nint GetHeight (EditorViewModel vm)
...
+		}
+
+		private void SetRowValueBackground (NSView view, bool valueBackground)
+		{
+			if (view == null)
+				return;
+
+			if (valueBackground) {
+				var c = this.hostResources.GetNamedColor (NamedResources.ValueBlockBackgroundColor);
+				view.SetValueForKey (c, new NSString ("backgroundColor"));
```
bleroy: No big deal, but make that a const?

ermau:
This also has some issues with the type selector, it can't be keyboard navigated properly.

vancura:
A couple of comments:

- Could you make the button standard? Not this outline one? Like here:

image

- Make the button wider, circa `20pt`.

- Did you have a chance to do the background treatment when editing values of that object?

mac object editor added 2x

ermau:
@vancura 

- Do you know what style that button is offhand?
- I can see if I can make it wider, I want to try to avoid a fixed size if I can.
- If you're talking about what I think you are, it should be visible in the second screenshot.

vancura:
> I can see if I can make it wider, I want to try to avoid a fixed size if I can.

Of course, but maybe it would be best if you could measure for the size and if it's less than `120pt`, set it to `120pt`. `Math.min(width, 120)` :)

> If you're talking about what I think you are, it should be visible in the second screenshot.

It is, indeed, and it looks good. Just one thing, could you grab a larger screenshot of the same thing for me? So I can measure? Here I can see just the top, and I'd like to see it in a bigger context. Also, how does it look in the dark theme, please?

ermau:
@vancura 
I don't think 120 is what you're looking for..
image

I swapped to the right bezel style, but it actually _breaks_ the popup behavior.

Dark mode (old bezel style, had to use to actually set a child)

image

Looks like type selector isn't quite right in dark theme though, I thought I had it correct but now it looks wrong, I'll add it to the list.

Light:
image


> Just one thing, could you grab a larger screenshot of the same thing for me? So I can measure?

Not sure what there's to measure past what's pictured, all rows are treated as normal rows, so any layout issues beyond that single object row are outside the scope of this PR.

vancura:
> I don't think 120 is what you're looking for..

Indeed, sorry. Could you use `70pt` instead, please?

ermau:
image

image

image


There is a remaining issue when the object is set and it's too wide, but I hope to come back and address that after #532.

bleroy:
If we're not supporting generic types, can we filter them out of the type picker? Currently, they trigger argument exceptions.

ermau:
> If we're not supporting generic types, can we filter them out of the type picker? Currently, they trigger argument exceptions.

I'd call that out of scope as it would equally apply to windows and be applied in the core.

Issue #518, opened 2019-02-01
Rectangle / Size / Point don't layout well on constrained sizes


ermau:
The spacing between the left and right sides almost seems fixed, it doesn't reduce it to try to fit when in a narrow layout.

CartBlanche:
@vancura With the Rectangle / Size / Point controls on Mac, I thought the spec was to have them fixed size. Should the resize as per control width. Do they have a maximum and minimum size, if so, what should that be?

vancura:
I wouldn't define any min or max, the panel itself helps in that: when the width is too low, all the controls look bad no matter what, and when it's really wide, all the controls should still look good (because of the contents alignment).

Also, if we say a control has a minimum width, what would happen when the panel is too narrow? Horizontal scroller?

CartBlanche:
OK. So I'll constrain the right side control to the `Right` of the of the Left control, so it stays the right distance. Just need to work out how to keep the X/Y etc labels constrained in the right places during resizes.

PR #521, opened 2019-02-06
Resize Point and Rectangle controls based on overall width.

CartBlanche:
Controls resized as per control width

Supposed to fix this - https://github.com/xamarin/Xamarin.PropertyEditing/issues/518

Narrow:
screen shot 2019-02-14 at 10 06 13

Wide:
screen shot 2019-02-14 at 10 04 12

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
@@ -20,7 +20,10 @@ namespace Xamarin.PropertyEditing.Mac
 		protected BasePointEditorControl (IHostResourceProvider hostResources)
 			: base (hostResources)
 		{
-			XLabel = new UnfocusableTextField ();
+			XLabel = new UnfocusableTextField {
+				Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize),
+				TranslatesAutoresizingMaskIntoConstraints = false,
+			};
 
 			XEditor = new NumericSpinEditor<T> (hostResources) {
 				BackgroundColor = NSColor.Clear,
@@ -28,7 +31,10 @@ namespace Xamarin.PropertyEditing.Mac
 			};
 			XEditor.ValueChanged += OnInputUpdated;
 
-			YLabel = new UnfocusableTextField ();
+			YLabel = new UnfocusableTextField {
+				Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize),
+				TranslatesAutoresizingMaskIntoConstraints = false,
+			};
 
 			YEditor = new NumericSpinEditor<T> (hostResources) {
 				BackgroundColor = NSColor.Clear,
@@ -42,11 +48,24 @@ namespace Xamarin.PropertyEditing.Mac
 			AddSubview (YEditor);
 
 			this.AddConstraints (new[] {
-				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Width, NSLayoutRelation.Equal, 1f, 90f),
+				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Top, 1f, 0f),
+				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Left, 1f, 0f),
+				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Right, NSLayoutRelation.Equal, YEditor,  NSLayoutAttribute.Left, 1f, -10f),
 				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, DefaultControlHeight),
 
-				NSLayoutConstraint.Create (YEditor, NSLayoutAttribute.Width, NSLayoutRelation.Equal, 1f, 90f),
+				NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.Top, NSLayoutRelation.Equal, XEditor,  NSLayoutAttribute.Bottom, 1f, -4f),
+				NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, DefaultControlHeight),
+
+				NSLayoutConstraint.Create (YEditor, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Top, 1f, 0f),
+				NSLayoutConstraint.Create (YEditor, NSLayoutAttribute.Right, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Right, 1f, -32f),
+				NSLayoutConstraint.Create (YEditor, NSLayoutAttribute.Width, NSLayoutRelation.Equal, XEditor,  NSLayoutAttribute.Width, 1f, 0f),
 				NSLayoutConstraint.Create (YEditor, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, DefaultControlHeight),
+
+				NSLayoutConstraint.Create (YLabel, NSLayoutAttribute.Top, NSLayoutRelation.Equal, XLabel,  NSLayoutAttribute.Top, 1f, 0f),
+				NSLayoutConstraint.Create (YLabel, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, DefaultControlHeight),
+
+				NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, XEditor.Subviews[0], NSLayoutAttribute.CenterX, 1f, 0),
+				NSLayoutConstraint.Create (YLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, YEditor.Subviews[0], NSLayoutAttribute.CenterX, 1f, 0),
 			});
 
 			ViewDidChangeEffectiveAppearance ();
--- a/Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
@@ -1,7 +1,8 @@
-﻿using System;
+using System;
 using System.Drawing;
 using AppKit;
 using CoreGraphics;
+using Foundation;
 using Xamarin.PropertyEditing.Drawing;
 using Xamarin.PropertyEditing.ViewModels;
 
@@ -14,26 +15,18 @@ namespace Xamarin.PropertyEditing.Mac
 			: base (hostResources)
 		{
 			// TODO localize
-			XLabel.Frame = new CGRect (34, 28, 25, 22);
-			XLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize); // TODO: Washed-out color following specs
 			XLabel.StringValue = "X";
 
 			XEditor.Frame = new CGRect (0, 46, 90, 20);
-
-			YLabel.Frame = new CGRect (166, 28, 25, 22);
-			YLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize); // TODO: Washed-out color following specs
+			
 			YLabel.StringValue = "Y";
 
 			YEditor.Frame = new CGRect (132, 46, 90, 20);
-
-			WidthLabel.Frame = new CGRect (20, -5, 50, 22);
-			WidthLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize); // TODO: Washed-out color following specs
+			
 			WidthLabel.StringValue = "WIDTH";
 
 			WidthEditor.Frame = new CGRect (0, 13, 90, 20);
-
-			HeightLabel.Frame = new CGRect (150, -5, 50, 22);
-			HeightLabel.Font = NSFont.FromFontName (DefaultFontName, DefaultDescriptionLabelFontSize); // TODO: Washed-out color following specs
+			
 			HeightLabel.StringValue = "HEIGHT";
 
 			HeightEditor.Frame = new CGRect (132, 13, 90, 20);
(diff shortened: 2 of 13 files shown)


CartBlanche:
@vancura Could you please take a look at the screenshot above.

ermau:
What does this look like when it's shrunk down small? Seems like we need a max size on the spacing rather than a fixed size.

CartBlanche:
@ermau There are 2 screenshot above. Once with it shrunk quite small and another larger one. Do you mean even smaller than that? We can reduce the spacing, but that will only gain us 50-100pixels.

ermau:
If the top image is representative of how this PR behaves, I definitely think that it should line up with the other controls like the second screenshot.

vancura:
No, @CartBlanche, sorry, this is not correct. See:

image

CartBlanche:
@vancura OK

CartBlanche:
@vancura Can you comment on the spacing between X & Y controls. Should the spacing be less now, between them? As in Right of X control, to Left of Y control.

CartBlanche:
Reduced spacing with narrowest width...
screen shot 2019-02-08 at 16 13 14

Still reduced spacing with wider width...
screen shot 2019-02-08 at 16 13 40

vancura:
It should be `18pt`:

image

Your gap seems to be `32pt`.

vancura:
Actually, sorry, can you make it `10pt`? So it matches other alignments.

CartBlanche:
@vancura Here is is a 10, according to Constraints...

screen shot 2019-02-11 at 13 53 20

vancura:
This is great, thanks!

BTW the right column is still here, but I think you mentioned you would fix it in another PR? Sorry, my memory is sparse.

CartBlanche:
@vancura Yes, right column removal will in in another PR.

vancura:
Is this the ticket where I can ask you to add two label tweaks?

Your version: 
image

My version:
image

Differences:
- [ ] Your label font size is `10pt`, make it `9pt`.
- [ ] Move the label `1pt` up (make the vertical gap between the textfield and label `1pt` smaller) – this might be unnecessary as the label font size might fix this, so let's see.

Thanks!

CartBlanche:
@vancura 9pt Font and 1 pt up. Any good?

screen shot 2019-02-11 at 15 14 40

CartBlanche:
@vancura To my view of symmetry, the right of the `Property Filter` search box, should really line up with the right edge of the controls, rather than the property button. It looks lop sided to me. Could just be me of course and that might be a tweak for another PR, if it is needed.

vancura:
@CartBlanche In fact, the whole header part was designed years ago, but never implemented. It's created to be globally used, not just in Property Pads.

The alignment will make more sense when the gap on the right is correct, I promise :)

CartBlanche:
@ermau if you have a minute look over the code changes. Should hopefully be straight forward.

ermau:
@vancura Those other measurements are unrelated to this PR, I'm assuming they're logged in the devops set of issues.

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
```diff
@@ -42,10 +48,19 @@ protected BasePointEditorControl (IHostResourceProvider hostResources)
 			AddSubview (YEditor);
 
 			this.AddConstraints (new[] {
-				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Width, NSLayoutRelation.Equal, 1f, 90f),
+				NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Top, 1f, 16f),
```
ermau: Shouldn't the labels constrain relative to what they're labeling? 

`NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.Top, NSLayoutRelation.Equal, XEditor,  NSLayoutAttribute.Bottom, 1f, 2),`

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
```diff
@@ -14,29 +14,28 @@ protected RectangleEditorControl (IHostResourceProvider hostResources)
...
+			
 			HeightLabel.StringValue = "HEIGHT";
 
 			HeightEditor.Frame = new CGRect (132, 13, 90, 20);
+
+			this.AddConstraints (new[] {
+				NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.Left, NSLayoutRelation.Equal, XEditor, NSLayoutAttribute.CenterX, 1f, -10f),
+				NSLayoutConstraint.Create (YLabel, NSLayoutAttribute.Left, NSLayoutRelation.Equal, YEditor, NSLayoutAttribute.CenterX, 1f, -10f),
+				NSLayoutConstraint.Create (WidthLabel, NSLayoutAttribute.Left, NSLayoutRelation.Equal, WidthEditor, NSLayoutAttribute.CenterX, 1f, -25f),
+				NSLayoutConstraint.Create (HeightLabel, NSLayoutAttribute.Left, NSLayoutRelation.Equal, HeightEditor, NSLayoutAttribute.CenterX, 1f, -25f),
```
ermau: You can specify `CenterX` on the `HeightLabel` so you don't need to do any manual adjustment that will assuredly break when the labels are allowed to be changed.

`NSLayoutConstraint.Create (HeightLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, HeightEditor, NSLayoutAttribute.CenterX, 1f, 0),`

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
ermau: Also why does this happen here when the other label constraints are added in the base?

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
```diff
@@ -42,10 +48,19 @@ protected BasePointEditorControl (IHostResourceProvider hostResources)
...
+				NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, DefaultControlHeight),
+
+				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Left, 1f, 0f),
+				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Right, NSLayoutRelation.Equal, YEditor,  NSLayoutAttribute.Left, 1f, -10f),
+				NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Width, 1f, -280f),
```
ermau: This seems like a wild guess and produces unusable results. Use the ratio argument to say "half the host, minus 10 for spacing" :
`NSLayoutConstraint.Create (XEditor, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Width, 0.5f, -10f),`

ermau:
image

On first load, the sizing of these controls isn't at all related with its host width. I think this will probably be addressed by some of the changes I suggested.

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
CartBlanche: No, because it will centre on the whole control, including spinners, which we don't want it to do, we just want it centred on the editing part of the control. At least that was the initial design from @vancura . If that has changed, then I'll change it to be centre on everything.

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
CartBlanche: IIRC this caused a constraints crash/collision, but I'll double check it.

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
ermau: You could include a manual offset for the spinners themselves, a negative offset of half the width of the spinners.

CartBlanche:
My first load, looks fine. Launch, click `Mock 2`. Give me this...

screen shot 2019-02-13 at 19 12 13

What are the steps to repro yours?

Xamarin.PropertyEditing.Mac/Controls/Layout.cs
```diff
@@ -3,6 +3,6 @@ namespace Xamarin.PropertyEditing.Mac
 {
 	internal static class Layout
 	{
-		internal const float GoldenRatioLeft = .38196601125f;
```
ermau: Why was this changed? This will affect a number of things and I'm not sure how you arrived at the new number, but I'm pretty sure it's incorrect.

ermau:
> My first load, looks fine. Launch, click Mock 2. Give me this...

It was reliable, now it's sporadic. I'll keep poking at it.

Xamarin.PropertyEditing.Mac/Controls/Layout.cs
CartBlanche: I was playing around with the left margin and forget to reset it back. Will be fixed in next push.

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
CartBlanche: The Labels were hard coded as `X`, `Y`, `WIDTH`, `HEIGHT` etc because we had a discussion about how they wouldn't change. Otherwise I would have put them in for localisation resources. We discussed this with Alan like a year or possibly more ago.

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
CartBlanche: Fixed.

Xamarin.PropertyEditing.Mac/Controls/BasePointEditorControl.cs
CartBlanche: That width shouldn't be there. Removed.

CartBlanche:
@ermau Please grab latest.

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
ermau: We discussed them not being localized which is not the same as not being changed, they are a property name and therefore subject to the platform. We haven't implemented the functionality yet, but they will absolutely have the option of changing in the future. Beyond that, your manual offset is still subject to the font and font size. Is there another technical concern with doing it this way?

CartBlanche:
@vancura Do the horizontal alignment on the Point/Rect etc control looks fine to you...
screen shot 2019-02-15 at 16 25 39

Xamarin.PropertyEditing.Mac/Controls/RectangleEditorControl.cs
CartBlanche: I think calculating the offsets, with font info will incur an overhead, but I've made the change. We'll find out if that is likely or not. Just awaiting Vaclav's approval that it looks fine for him.

ermau: You're right, that would incur an overhead. Was there an issue with my suggested solution?

CartBlanche: @ermau  Do you mean this...
> You could include a manual offset for the spinners themselves, a negative offset of half the width of the spinners.

?

ermau: > Do you mean this?

That and
`NSLayoutConstraint.Create (HeightLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, HeightEditor, NSLayoutAttribute.CenterX, 1f, 0),`

CartBlanche: `NSLayoutConstraint.Create (HeightLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, HeightEditor, NSLayoutAttribute.CenterX, 1f, 0),` will not take into account the width of the actual text in the label, it will position it exactly in the centre of the label. Adding the spinner offsets, unfortunately isn't enough for it to position the text correctly. 

As allowing custom labels in on the road-map, doing the width calculations is the best way to ensure we handle things correctly, regardless of font and the text users might use. Please take a look at the calculations. There might be another way to minimise the calculations overheads. It doesn't look massive right now, but might be worth another pair of eyes.

ermau: > will not take into account the width of the actual text in the label, it will position it exactly in the centre of the label.

The label already sizes to text, how would the center of the label not also be the center of the text?
image

I checked out your branch and tested this change to the thickness editor (we can actually skip the spinner offset by constraint directly to the textbox):
```csharp
this.AddConstraints (new[] {
	NSLayoutConstraint.Create (XLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, XEditor.Subviews[0], NSLayoutAttribute.CenterX, 1f, 0),
	NSLayoutConstraint.Create (YLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, YEditor.Subviews[0], NSLayoutAttribute.CenterX, 1f, 0),
	NSLayoutConstraint.Create (WidthLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, WidthEditor.Subviews[0], NSLayoutAttribute.CenterX, 1f, 0),
	NSLayoutConstraint.Create (HeightLabel, NSLayoutAttribute.CenterX, NSLayoutRelation.Equal, HeightEditor.Subviews[0], NSLayoutAttribute.CenterX, 1f, 0),
});
```

image

As you can measure, the thickness labels are centered perfectly (give or take 1px because the widths aren't even) on the textbox above it.

CartBlanche: I hadn't thought of using `SubViews[0]`. That works really well and simplifies the code across the board and no calcs on our part.

CartBlanche: Done.

PR #528, opened 2019-02-11
[Core] Fix grouped type properties not swapping properly

### LINKED ISSUE #525: Crash adding item to collection with no properties
1. Add a suggested type that has properties
2. Add an instance of the type from 1.
3. Add a suggested type that has no properties
4. Add an instance of the type from 3.
Boom.

ermau:
Fixes #525

When we moved to the new panel group system, the grouped editors weren't
accounted for and so when we try to remove their child property from the
container group it's not found since they're inside the group editor VM.

This also meant that even if removal had worked, we wouldn't properly
add new shared view model types to the group.

This version addresses the crasher listed in #527, the second issue was a false alarm.

### DIFF
--- a/Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
+++ b/Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
@@ -10,11 +10,14 @@ namespace Xamarin.PropertyEditing.ViewModels
 	internal class PanelGroupViewModel
 		: NotifyingObject
 	{
-		public PanelGroupViewModel (string category, IEnumerable<EditorViewModel> editors, bool separateUncommon = true)
+		public PanelGroupViewModel (TargetPlatform targetPlatform, string category, IEnumerable<EditorViewModel> editors, bool separateUncommon = true)
 		{
+			if (targetPlatform == null)
+				throw new ArgumentNullException (nameof(targetPlatform));
 			if (editors == null)
 				throw new ArgumentNullException (nameof(editors));
 
+			this.targetPlatform = targetPlatform;
 			Category = category;
 			AddCore (editors, separateUncommon);
 		}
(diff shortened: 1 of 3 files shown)


Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
```diff
@@ -246,19 +271,20 @@ protected override void OnAddEditors (IEnumerable<EditorViewModel> editors)
...
-				if (remainingItems != null) // TODO: pretty sure this was out of order before, add test
-					this.arranged.Add (key, new PanelGroupViewModel (key, grouping.Where (evm => remainingItems.Contains (evm))));
-				else
-					this.arranged.Add (key, new PanelGroupViewModel (key, grouping, separateUncommon: !isFlat));
+				string key = grouping.Key;
+				if (remainingItems != null) {// TODO: pretty sure this was out of order before, add test
+					if (remainingItems.Count > 0)
+						this.arranged.Add (key, new PanelGroupViewModel (TargetPlatform, key, grouping.Where (evm => remainingItems.Contains (evm))));
+				} else
+					this.arranged.Add (key, new PanelGroupViewModel (TargetPlatform, key, grouping, separateUncommon: !isFlat));
```
bleroy: Not from this change, but it's unclear why we care about `separateUncommon` in one case and not the other. Might want to add a comment when you get the chance.

Xamarin.PropertyEditing/ViewModels/PropertyGroupViewModel.cs
```diff
@@ -55,10 +55,35 @@ public string FilterText
 			}
 		}
 
-		private readonly PropertyViewModel[] properties;
+		public void Add (PropertyViewModel property)
+		{
+			if (property == null)
+				throw new ArgumentNullException (nameof(property));
+
+			this.properties.Add (property);
```
bleroy: So why is it not added to `propertiesView`?

ermau: `Reload()` handles that.

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
ermau: In the first case we can assume we're in a grouped mode as there wouldn't be remaining items otherwise. We could pass the same `!isFlat` to make things clearer I suppose, it would always be `true`.

Issue #486, opened 2018-12-15
The texts of the labels are cut into standard window sizes


netonjm:
![image](https://user-images.githubusercontent.com/1587480/50037499-54956900-0012-11e9-87b9-2704d3c6643c.png)

CartBlanche:
As a temporary stop gap, would setting tooltips, like this help...
<img width="333" alt="screen shot 2019-02-11 at 09 50 04" src="https://user-images.githubusercontent.com/271363/52555760-8ba1b400-2de2-11e9-8026-41a91f04a582.png">

Eric mentioned on a call last week that ultimately, tooltips will hold description information too.

I am investigating a better solution, but don't want this sort of thing holding up the release or port to Vs4Mac.

netonjm:
Maybe @vancura can help here too

vancura:
This was designed as the first thing ever:

![image](https://user-images.githubusercontent.com/4982/52561109-ba2b8900-2dfa-11e9-92fe-02d8f7c6b7ca.png)

alanmcgovern:
What if it's not multiline?

alanmcgovern:
Specifically, in the XAML case we show the name of the C# property as the label. I took a quick scout for a long one and this showed up: `HorizontalScrollBarVisibility`

vancura:
I believe we, as MSFT, should keep the voice human and not robotic (see https://medium.com/microsoft-design/microsoft-voice-principles-1-730f413190c1). This is precisely the place where we easily sound like robots. Is there a way to translate IDs like `HorizontalScrollBarVisibility` to `Horizontal scroll bar visibility:`?

alanmcgovern:
In this scenario the name in the property panel is a 1:1 match with the name used in the C# property declaration and also the XAML file itself. This is the same behaviour as WPF/UWP.

We could create a file, or annotation, for every public property in Xamarin.Forms but it would be an expensive process.

CartBlanche:
As per my screenshot above, the long property names are definitely not multi-line.
I think the tooltip, though not brilliantly aesthetic, can be made to only appear for properties that have been cut off (the code for that is already working). Ideally if we can reposition the tooltip's popup, so it happens right over the cut off text, then that is the kind of behaviour we have on Windows, IIRC.

hbons:
Some macOS apps also squeeze the letter spacing when space is tight. It that a possibility?

CartBlanche:
[...]

PR #529, opened 2019-02-11
[Mac] Set Tooltip value for when resizing occurs & long properties are…

CartBlanche:
… clipped

Fixes: https://github.com/xamarin/Xamarin.PropertyEditing/issues/486

I did try hooking into 
		```
public override void ResizeWithOldSuperviewSize (CGSize oldSize)
		{
			base.ResizeWithOldSuperviewSize (oldSize);
		}
```

and tracking `ObserveFrameChanged`

But both of these fire for all controls in the property list, so make resizing very laggy and almost unusable. So pulled that code out.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
@@ -1,5 +1,7 @@
 using System;
 using AppKit;
+using CoreGraphics;
+using Foundation;
 
 namespace Xamarin.PropertyEditing.Mac
 {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by CartBlanche** [comment]
This is a temporary fix and the long term goal will be to investigate how we can make sure the UnforcusableTextField control can detect it is being clipped. Ideally triggered from the EditorContainer letting it know.

File: Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs [hunk 1]
```diff
@@ -55,5 +60,6 @@ public EditorContainer (IHostResourceProvider hostResources, IEditorView editorV
 #endif
 
 		private readonly IHostResourceProvider hostResources;
+		private nfloat labelToControlSpacing = 5f;
```
ermau: Seems like this should be `const`?

CartBlanche: Yes was meant to be.

Issue #531, opened 2019-02-13
[Mac] String and Numeric Controls aren't handling clipped values.


CartBlanche:
As per this StringControl example.

<img width="246" alt="screen shot 2019-02-13 at 11 02 43" src="https://user-images.githubusercontent.com/271363/52718638-d9155100-2f9b-11e9-8664-6401a279638e.png">

PR #532, opened 2019-02-13
[Mac] Add new BasePropertyControlTextField to have ellipses and Allow…

CartBlanche:
…sExpansionToolTips by default.

Fixes: #531 

screen shot 2019-02-13 at 14 53 39

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using AppKit;
 using CoreGraphics;
 using Foundation;
@@ -6,13 +6,13 @@ using ObjCRuntime;
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	public class NumericTextField : NSTextField
+	internal class NumericTextField : PropertyTextField
 	{
-		NSText CachedCurrentEditor {
+		private NSText CachedCurrentEditor {
 			get; set;
 		}
 
-		string cachedValueString;
+		private string cachedValueString;
 
 		public bool AllowNegativeValues {
 			get; set;
@@ -54,7 +54,7 @@ namespace Xamarin.PropertyEditing.Mac
 		public override bool ShouldBeginEditing (NSText textObject)
 		{
 			CachedCurrentEditor = textObject;
-			cachedValueString = textObject.Value;
+			this.cachedValueString = textObject.Value;
 
 			if (AllowRatios)
 				CachedCurrentEditor.Delegate = new RatioValidateDelegate (this);
@@ -81,7 +81,7 @@ namespace Xamarin.PropertyEditing.Mac
 
 		public virtual void ResetInvalidInput ()
 		{
-			this.StringValue = cachedValueString;
+			this.StringValue = this.cachedValueString;
 		}
 
 		public static bool CheckIfNumber (string finalString, ValidationType mode, bool allowNegativeValues)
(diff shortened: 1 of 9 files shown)


Xamarin.PropertyEditing.Mac/Controls/Custom/NumericTextField.cs
```diff
@@ -93,9 +93,8 @@ public static bool CheckIfNumber (string finalString, ValidationType mode, bool
 
 		public static bool ValidateDecimal (string finalString, bool allowNegativeValues)
 		{
-			double value;
 			//Checks parsing to number
-			if (!double.TryParse (finalString, System.Globalization.NumberStyles.Any, System.Globalization.CultureInfo.CurrentUICulture, out value))
+			if (!double.TryParse (finalString, System.Globalization.NumberStyles.Any, System.Globalization.CultureInfo.CurrentUICulture, out var value))
```
ermau: Going forward, let's not make changes like this unless we're actually touching this line, we're just making the real history of this line harder to find and making it harder to review for nearly no benefit.

Xamarin.PropertyEditing.Mac/Controls/Custom/BaseTextField.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public class BasePropertyControlTextField : NSTextField
```
ermau: - Does this need to be public?
- Can we try a different name? It's not purely a base if we're using it directly.
- The file name really should match the class name

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
```diff
@@ -59,14 +59,10 @@ public UnfocusableTextField (CGRect frameRect, string text) : base (frameRect)
...
 		{
-			this.label = new NSTextField {
+			this.label = new BasePropertyControlTextField {
 				AccessibilityElement = false,
 				BackgroundColor = NSColor.Clear,
 				Bordered = false,
-				Cell = {
-					LineBreakMode = NSLineBreakMode.TruncatingTail,
-					UsesSingleLineMode = true,
-				},
```
ermau: Shouldn't this have meant the property editor labels would have had a ... and all that? They weren't and still don't, but should.

CartBlanche: They do, when clipped "correctly". Look at the `Type` field in the header Panel, when you resize to a narrow width. The `...` is there.

Xamarin.PropertyEditing.Mac/Controls/Custom/BaseTextField.cs
CartBlanche: Do you have any naming suggestions?
Only other name I can suggest would be one denoting functionality. 
So `EllipsisTextField` ?

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
ermau: So we need to fix the clipping separately? 👍

Xamarin.PropertyEditing.Mac/Controls/Custom/BaseTextField.cs
ermau: Could just be `PropertyTextField`

CartBlanche: Sure, why not.

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableTextField.cs
CartBlanche: Yes. Please see my comment on https://github.com/xamarin/Xamarin.PropertyEditing/pull/529

CartBlanche:
@ermau Changes resolved. New control renamed.

CartBlanche:
@ermau Who would have thought adding constraints would break keyboard navigation. If this is the only thing holding back this PR, I'll revert the change for this control and add an issue specifically to investigate why switching to constraints breaks navigation in this control.

ermau:
Just needs a rebase and good to go.

PR #533, opened 2019-02-13
[Mac] Tweak the edit rect and position and size of the control.

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/Custom/TextFieldSmallButtonContainer.cs
```diff
@@ -131,10 +131,12 @@ public override void ResetCursorRect (CGRect cellFrame, NSView inView)
 			public override CGRect DrawingRectForBounds (CGRect theRect)
 			{
 				CGRect baseRect = base.DrawingRectForBounds (theRect);
-				if (this.field.buttons.Count == 0) {
-					return baseRect;
+				if (this.field.buttons.Count != 0) {
+					baseRect.Y -= 2;
+					baseRect.Width -= this.field.ButtonSize;
 				}
-				return new CGRect (baseRect.X, baseRect.Y - 2, baseRect.Width + DefaultMargin - (this.field.ButtonSize * this.field.buttons.Count), DefaultHeight);
```
CartBlanche: No need to create totally new object here, if we can reuse what we have. Should be marginally faster.

Issue #536, opened 2019-02-15
Object editor should filter out generic types


bleroy:
Currently, if you pick a generic type, we throw an argument exception.

ermau:
The trick here is our type interface has no generic information so the best we could do is look for ```n` to filter them out. The reality is though that this was overlooked and it's actually valid to be able to select a generic type here, the trick is we'd need a way of selecting the type arguments.

In the interim it's my suggestion that the platform temporarily self-filter out generics.

PR #537, opened 2019-02-19
[Mac] Collection Editor

ermau:
screen shot 2019-03-07 at 15 55 34
Screen Shot 2019-03-12 at 18 28 00
Screen Shot 2019-03-12 at 18 28 12


vancura:
@ermau Could you please show me a screenshot on how this looks when the focus rectangle is out of the dropdown? Hard to see the metrics with it. Also, a dark theme screenshot would be super-helpful. Thanks!

vancura:
A few more things:

- [ ] Could you please make the _Edit_ button height by `3pt` smaller? It's too tall.
- [ ] Border around window should be standard `20pt`, you have `10pt`.
- [ ] Left table header height should be `21pt`, you have `19pt` (also watch out for the vertical label alignment, now it's too low).
- [ ] Left table row height should be `24pt`, you have `18pt`.
- [ ] Buttons should be (constraints minimum at least) `80pt`, you have `69pt`.
- [ ] Vertical gap between buttons and both tables should be `20pt`, you have `11pt`.

Thanks!

metrics

vancura:
One last thing, `Collection Editor: %s` sounds like designed for robots or people who know what an internal name _Collection Editor_ stands for. Please set the window title to `Edit Collection \u2013 %s` (the Unicode 2013 code stands for the dash, more info at https://gist.github.com/vancura/27fad9793840ceabae14#-en-dash).

ermau:
> a dark theme screenshot would be super-helpful. 

There's no dark theme screenshot because the dark theme colors are not present, see the design ticket in D3.

> Collection Editor: %s sounds like designed for robots or people who know what the internal name Collection Editor stands for

Just because we use the word "editor" internally doesn't make it an internal title, it actually comes from VS. If you still feel strongly I can change it, but it's a perfectly valid title.

hbons:
I changed the title intentionally. On the Mac side we try to avoid names like "editor", "manager", "explorer" etc. These are words that don't add much value and are kind of "meta". And agree with @vancura it's a bit robotic. :)

vancura:
@ermau: Could you please attach a few screenshots so I can review? I don't have time to compile and hunt for all possible scenarios.

ermau:
> @ermau: Could you please attach a few screenshots so I can review? I don't have time to compile and hunt for all possible scenarios.

@vancura I updated the PR description with new screenshots before requesting a rereview.

vancura:
Sorry, this is still not right. As with my previous review, the left is my design, and right is your current implementation:

image

- [ ] Padding between content and the left, right and top edge should be `20pt`, it's still `10pt`.
- [ ] The same applies to the button set.

Is it possible to tweak this space, marked with a red rectangle?

image

I'd like to have this space `8pt`, right now it seems to be `3pt` (hard to tell due to typography tricks of SF fonts).

Also, we don't have a chance to display the icons in the table?

vancura:
Also:

- [ ] Buttons need to be _Cancel_ and _Select_ (not _Cancel_ and _OK_), and _Select_ needs to be default (blue).

ermau:
> Buttons need to be Cancel and Select (not Cancel and OK)

"Select" doesn't make any sense in this scenario? What are you selecting?

> Select needs to be default (blue).

This will have to wait, there's a bug preventing me from enabling that, the whole thing doesn't need to block on it.

ermau:
> Padding between content and the left, right and top edge should be 20pt, it's still 10pt.

Yup, misread the correction image, will address.

> I'd like to have this space 8pt, right now it seems to be 3pt (hard to tell due to typography tricks of SF fonts).

Should be doable.

> Also, we don't have a chance to display the icons in the table?

The icons are hidden while the implementation doesn't provide an icon provider (as none of them currently do). This is to avoid just a sea of generic icons.

ermau:
> FYI, I only looked at the code since @vancura did a thorough review of the runtime behavior of that changeset.

To be clear, he's just reviewing my screenshots, not checking out and testing.

bleroy:
Oh ok, I'll check it out then.

bleroy:
The new element dropdown's behavior looks super-weird to me: you click the down arrow (I would expect to see the list of available types directly when I do this), then have to select "<other type>" in order to see the type selection popup.

bleroy:
I think he filter textbox for types of new elements should have a clear button.

bleroy:
Select a generic type, click '+' -> boom, exception.

bleroy:
The natural width of the collection editor makes the solid color picker look very narrow. In fact, most labels are clipped with the default size. Definitely too narrow by default.

bleroy:
When you edit a collection, select an item, close, then re-open, the latest selection remains, but is not visible (you can't see which item is selected).

ermau:
> Select a generic type, click '+' -> boom, exception.

This is provided from the same APIs as #536, so for right now it's up to the platform to not list generic types as "assignable".

> The natural width of the collection editor makes the solid color picker look very narrow.

The designed color picker is too wide, I've asked Vaclav to take another look at it. I can bump it up a bit, but it's already larger than I suspect most people will have their property panel expanded to anyway.

> The new element dropdown's behavior looks super-weird to me: you click the down arrow (I would expect to see the list of available types directly when I do this), then have to select "" in order to see the type selection popup.

The _initial_ list of types is controlled by https://github.com/xamarin/Xamarin.PropertyEditing/blob/master/Xamarin.PropertyEditing/AssignableTypesResult.cs#L28, the standalone doesn't have any suggested types. The option that you select to popup the type selector should have some text "<Other type...>"

> When you edit a collection, select an item, close, then re-open, the latest selection remains, but is not visible (you can't see which item is selected).

As in you can still see the properties? Can you be sure you've actually fetched and checked out `origin/ermau-mac-collectioneditor`? This should have been addressed by d9e7bd9

> I think he filter textbox for types of new elements should have a clear button.

This is a shared component, can you make a separate issue for that?

bleroy:
It's not just the color picker that's too narrow, other properties are also so narrow the labels get cropped.

Yes, I can still see the properties of the last selection, but in the list, it doesn't appear as selected. And yes, I checked out the latest from that branch.

Filter clear selection: https://github.com/xamarin/Xamarin.PropertyEditing/issues/559

ermau:
> Yes, I can still see the properties of the last selection, but in the list, it doesn't appear as selected. And yes, I checked out the latest from that branch.

Ah yeah, that was a separate issue. I've fixed a leak and that issue.

PR #542, opened 2019-02-21
Ensure Source can only be set via constructor.

CartBlanche:


Xamarin.PropertyEditing/Common/FilePath.cs
```diff
@@ -38,7 +36,7 @@ public override int GetHashCode ()
 		{
 			var hashCode = 1861433795;
```
netonjm: could you please include this like a constant in the class? I forgot about change it

CartBlanche: @netonjm Where would it be needed again? In the property editor we haven't used `const` fr `hashCode` anywhere.

netonjm: This integer generates an allocation for this integer every time you call to GetHasCode, which is unnecessary because doesn't change in time.

ermau: I think he's asking where _else_ would it be needed for it to be a _class_ const. You could just make it a local const.

netonjm: exactly, thanks @ermau !

CartBlanche: a local `const` would make sense, but I think this was avoided in other `GetHashCode` functions because we'd then need to declare another variable for us to then return from the function.

If we want to go down that route, may I suggest we raise an issue and do a pass through the code base to make it standard practice where-ever we are calculating hashcodes.

Therzok: Yeah, constant folding is done if it's a `const int` or `const var` rather than just `int`.

ermau: It's fine as-is.

PR #543, opened 2019-02-22
Enabled must be set for ReadOnly to look and behave correctly. Als…

### LINKED ISSUE #499: [Mac] No visual difference between not editable/editable text properties
Text properties should have a visual difference with respect to the editable ones.

readonlytextfield

It should have similar behavior we have with the checkboxes.

image

CartBlanche:
…o fixes InputMode enablement. Also turned out that `CombinablePropertyEditor` wasn't setting the checkboxes to `ReadOnly` either.

Fixes #499 

screen shot 2019-02-27 at 17 53 04

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/StringEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/StringEditorControl.cs
@@ -25,6 +25,8 @@ namespace Xamarin.PropertyEditing.Mac
 		internal NSPopUpButton inputModePopup;
 		private IReadOnlyList<InputMode> viewModelInputModes;
 
+		private bool CanEnable => ViewModel.Property.CanWrite && (((ViewModel.InputMode != null) && !ViewModel.InputMode.IsSingleValue) || (this.inputModePopup == null));
+
 		public StringEditorControl (IHostResourceProvider hostResource)
 			: base (hostResource)
 		{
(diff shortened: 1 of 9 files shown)


Xamarin.PropertyEditing.Mac/Controls/StringEditorControl.cs
```diff
@@ -120,7 +122,7 @@ protected override void UpdateErrorsDisplayed (IEnumerable errors)
 
 		protected override void SetEnabled ()
 		{
-			this.stringEditor.Editable = ViewModel.Property.CanWrite;
+			this.stringEditor.Editable = this.stringEditor.Enabled = ViewModel.Property.CanWrite;
```
netonjm: The Enabled property with False value makes the field **not focusable** and therefore not editable, maybe then we can remove it.

CartBlanche: If we remove the Enabled then we don't get the text changing to light gray automagically. Which means we'd need to create a custom TextField, to change the TextColour depending on Dark/Light mode when enabled or disabled.
What scenario do you feel needs you to be able to focus a disabled control?

Note: None of the numeric editors allow focusing when they are disabled, nor the drop-downs or popups.

CartBlanche: Also Xcode behaves the same way. If a `NSTextField` is disabled, when tabbing it will be skipped. So I think this is expected behaviour on Mac.

netonjm: I mean remove the Editable property set, because with a Enabled field in disabled user cannot edit

CartBlanche: Yes. Thanks. Fixed.

CartBlanche:
Not sure what is going on with Numeric Fields. The SpinnerButtons are obviously disabled, and you can't interact with it, but oddly the text within isn't disabled. So enabled flag is being set. Investigating.

CartBlanche:
Fixed Enablement. Broken other stuff.

CartBlanche:
Enablement should work across the board now.

Issue #466, opened 2018-12-03
[Mac] When tabbing, focused control should scroll into view.


CartBlanche:
Steps:
1. Start Sample on Mac
2. Select a Control in the Property Panel so it has focus.
3. Tab through controls

Currently:
When you tab out of the visible area it focuses on a control, but it is out of view.

Expected:
When you tab to a control that is out of the visiable area, it should scroll that control into view.

PR #545, opened 2019-02-25
[Mac] When tabbing scroll focused control into view.

CartBlanche:
Fixes https://github.com/xamarin/Xamarin.PropertyEditing/issues/466

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BooleanEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BooleanEditorControl.cs
@@ -14,14 +14,7 @@ namespace Xamarin.PropertyEditing.Mac
 		public BooleanEditorControl (IHostResourceProvider hostResource)
 			: base (hostResource)
 		{
-			BooleanEditor = new NSButton {
-				AllowsMixedState = true,
-				ControlSize = NSControlSize.Small,
-				Font = NSFont.FromFontName (DefaultFontName, DefaultFontSize),
-				Title = string.Empty,
-				TranslatesAutoresizingMaskIntoConstraints = false,
-			};
-			BooleanEditor.SetButtonType (NSButtonType.Switch);
+			BooleanEditor = new FocusableBooleanButton ();
 
 			// update the value on 'enter'
 			BooleanEditor.Activated += (sender, e) => {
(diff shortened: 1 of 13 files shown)


Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyTextField.cs
```diff
@@ -11,5 +11,11 @@ public PropertyTextField ()
 			Cell.LineBreakMode = NSLineBreakMode.TruncatingTail;
 			Cell.UsesSingleLineMode = true;
 		}
+
+		public override bool BecomeFirstResponder ()
+		{
+			ScrollRectToVisible (this.Bounds);
```
ermau: We don't need `this` here and we should only actually scroll if `base.Become..` returns true.

CartBlanche: From testing `BecomeFirstResponder ()` is only called when the control is about to receive focus. What scenarios have you seen where `BecomeFirstResponder ()` is called and our property controls should *not* scroll into view? @alanmcgovern Any insights here? I'm happy to make the change, but want to understand exactly why the check is absolutely necessary.

ermau: If you read the docs on `BecomeFirstResponder`, you'll see that the return value is whether or not it actually became first responder. If it hasn't actually become first responder, we don't want to scroll to it. It doesn't matter whether we've seen it happen or not, implementing it correctly is a defensive measure.

CartBlanche: I had already changed it, but hadn't pushed. Done.

CartBlanche:
I had a change for Bool/Combinable, but had not pushed it. I was going to ask about Object/Type though. So I'll just make those accept keyboard focus and that's it then.

CartBlanche:
All controls are now focusable. But I'm now investigating ghost focusing of Brush control, when it is ReadOnly.

CartBlanche:
Ok, so ensuring CanBecomeKeyView is linked to Enabled give the correct behaviour.

PR #549, opened 2019-02-28
[build] Ensure the pngs are copied where we expect

alanmcgovern:
This makes sure they end up in the right directory

### DIFF
--- a/Xamarin.PropertyEditing.Mac.Standalone/Xamarin.PropertyEditing.Mac.Standalone.csproj
+++ b/Xamarin.PropertyEditing.Mac.Standalone/Xamarin.PropertyEditing.Mac.Standalone.csproj
@@ -114,19 +114,13 @@
     </ProjectReference>
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\Mac\Xamarin.Mac.CSharp.targets" />
-  <ItemGroup>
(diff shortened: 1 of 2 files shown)


alanmcgovern:
These are the places where we end up copying the icon set to:
```
Xamarin.PropertyEditing.Mac/bin/Debug/PropertyEditingResource.bundle
Xamarin.PropertyEditing.Mac.Standalone/obj/Debug/actool/bundle
Xamarin.PropertyEditing.Mac.Standalone/bin/Debug/Xamarin.PropertyEditing.Mac.Standalone.app/Contents/MonoBundle/PropertyEditingResource.bundle
Xamarin.PropertyEditing.Mac.Standalone/bin/Debug/PropertyEditingResource.bundle
```
This looks good as it's in the right location in the appbundle and was appropriately local copied to the output directory of the project which references `Xamarin.PropertyEditing.Mac`

CartBlanche:
@alanmcgovern Looks good and works for Standalone. So will that also work once we bump Jose's PR (https://github.com/mono/monodevelop/pull/6742) for him, based on the previous build changes?

alanmcgovern:
Yup! It should make the changes in jose's branch unnecessary.

The only gotcha is that Xamarin.Mac based projects do not take transitive `Content` files into account when generating the final appbundle, which they should. That's why xamarin.mac executable projects still need to manually copy the bundle from their output dir inside the app bundle dir.

PR #558, opened 2019-03-11
Fix filtering issues

ermau:


Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
```diff
@@ -68,7 +69,13 @@ public bool Remove (EditorViewModel editor)
 				}
 			}
 
-			return list.Remove (editor);
+			bool result = list.Remove (editor);
+			if (result) {
+				OnPropertyChanged (nameof(HasChildElements));
```
bleroy: Inconsistent spacing.

PR #560, opened 2019-03-12
[Mac] Introduce a version of the mac control where the host resource provider can be passed into the constructor

bleroy:
The resources are consumed in Initialize, which is called by the constructors, so we need a way to set it as part of construction.


ermau:
Should we be including the move default colors commit if it didn't help anything?

bleroy:
You're right, I'll remove it.

bleroy:
Done and rebased.

Issue #284, opened 2018-04-17
Mac Binding Editor


ermau:

vancura:
Design and talk at https://devdiv-design.visualstudio.com/D3%20Studio/_workitems/edit/3588.

PR #561, opened 2019-03-14
[Mac] Initial Binding Editor Dialog.

CartBlanche:
Fixes: https://github.com/xamarin/Xamarin.PropertyEditing/issues/284


Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,274 @@
...
+					NSLayoutConstraint.Create (label, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, 18f),
+				});
+
+				var textField = new NSTextField {
+					Enabled = item.Property.CanWrite,
+					StringValue = item.Name,
+					TranslatesAutoresizingMaskIntoConstraints = false,
+				};
+				textField.Changed += (sender, e) => {
+					// TODO What do we update in the VM with this value?
```
CartBlanche: @ermau I'm finding it difficult to pin point exactly which property on the VM this should hook up to. Any pointers?

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,274 @@
...
+					ControlSize = NSControlSize.Small,
+					Enabled = item.Property.CanWrite,
+					Font = NSFont.FromFontName (PropertyEditorControl.DefaultFontName, PropertyEditorControl.DefaultFontSize),
+					Title = item.Property.Name,
+					TranslatesAutoresizingMaskIntoConstraints = false,
+				};
+
+				flag.SetButtonType (NSButtonType.Switch);
+				flag.Activated += (sender, e) => {
+					// TODO What do we update in the VM with this value?
```
CartBlanche: @ermau Same here. I'm finding it difficult to pin point exactly which property on the VM this should hook up to. Any pointers?

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
ermau: It shouldn't even be a text field, it should be a full on property presenter.

CartBlanche: @ermau I don't have notes saying that section should have specifically been a property presenter, only that that section was dynamic based on the `BindingProperties` and `FlagProperties`. Could you clarify where/when that was mentioned, I must have missed that?

So were you thinking that we should re-use `EditorContainer` or did you have something else in mind or am I totally off track here?

ermau: I don't know what else you would do, why would you expect to take random `PropertyViewModel`s from `BindingProperties` and make a new editor for them here instead of reusing everything we've already built?

Yes, we should reuse `EditorContainer`.

CartBlanche:
Vaclav please cast your design eye over these screenshots...
Screen Shot 2019-03-20 at 15 59 05
Screen Shot 2019-03-20 at 15 59 36
Screen Shot 2019-03-20 at 15 59 58

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/CreateValueConverterWindow.cs
```diff
@@ -0,0 +1,125 @@
...
+				NSLayoutConstraint.Create (buttonDone, NSLayoutAttribute.Height, NSLayoutRelation.Equal, 1f, 24),
+			});
+
+			ContentViewController = new NSViewController (null, null) {
+				View = container,
+			};
+
+			ViewModel.PropertyChanged += (sender, e) => {
+				if (e.PropertyName == nameof (ViewModel.SelectedType)) {
+					this.valueConverterName.StringValue = ViewModel.SelectedType.Name;
```
alanmcgovern: This will be null if you select something in the treeview and then click on an empty space to unselect it.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -228,5 +243,11 @@ private void OnResourceRequested (object sender, EventArgs e)
 
 			resourceSelectorPopOver.Show (requestResourceView.Frame, (NSView)this, NSRectEdge.MinYEdge);
 		}
+
+		private void OnBindingRequested (object sender, EventArgs e)
+		{
+			var bindingEditorWindow = new BindingEditorWindow (this.hostResources, this.viewModel);
+			bindingEditorWindow.MakeKeyAndOrderFront (this);
```
alanmcgovern: Should this be run modally too? There's nothing stopping someone from opening the binding editor, then clicking in the panel and opening a second one

vancura:
A couple of comments:

image

- [ ] Please use different space around the window, your is too low
- [ ] Also space between top dropdown section and two tables
- [ ] The Path / Custom checkbox need to be centered
- [ ] Left panel has no header (in my comp it writes "Ancestor Type and Level")
- [ ] The list on the right (your "MockSampleControl") shouldn't have a border around, it should be edge to edge
- [ ] Space between two panels and the Converter section is too low
- [ ] Other Settings should be always named Other Settings, even when it's expanded (you have Hide Settings in that case)
- [ ] Please vertically center the Other Settings panel contents, it's too much on top.

vancura:
image

(left my design, right your screenshot)

- [ ] Different padding around the window
- [ ] Filter and the MockValueConverter don't horizontally align
- [ ] Done button shouldn't be aligned with the Show all assemblies checkbox
- [ ] Show All Assemblies should be capitalized
- [ ] Why only Done button?

Do you show the helper text when there's nothing listed?

ermau:
> Left panel has no header (in my comp it writes "Ancestor Type and Level")
It actually does have a header, you can see it in the second screenshot. It's dependent on the selected binding type.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingPathSelectorControl.cs
```diff
@@ -0,0 +1,339 @@
...
+			AddSubview (customCheckBox);
+			AddConstraints (new[] {
+				NSLayoutConstraint.Create (customCheckBox, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this, NSLayoutAttribute.Top, 1f, 8f),
+				NSLayoutConstraint.Create (customCheckBox, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this, NSLayoutAttribute.Right, 1f, -90f),
+			});
+
+			this.CustomPath = new NSTextField {
+				ControlSize = NSControlSize.Mini,
+				Enabled = false,
+				Font = NSFont.FromFontName (PropertyEditorControl.DefaultFontName, PropertyEditorControl.DefaultFontSize),
```
ermau: Are we sure `Mini` is the right size here? The text sits at the top of what looks like a full or at least small size control in the actual window (as it's using `DefaultControlHeight` which matches what it'd be for `Small`: 18.) These properties are also effectively deprecated, if it's `Mini` or `Small`, use the system method for getting the correct size (which `DefaultFontSize` isn't correct for `Mini`).

```suggestion
				Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Mini)),
```

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingTypeSelectorControl.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.ComponentModel;
+using AppKit;
+using Foundation;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class BindingTypeSelectorControl : NSView
```
ermau: Do we need to include this implementation? Doesn't appear to be used.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/CreateValueConverterWindow.cs
ermau: This was marked resolved but wasn't actually fixed.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,276 @@
...
+		internal CreateBindingViewModel ViewModel => this.viewModel;
+
+		internal BindingEditorWindow (IHostResourceProvider hostResources, PropertyViewModel propertyViewModel) 
+		{
+			this.viewModel = new CreateBindingViewModel (propertyViewModel.TargetPlatform, propertyViewModel.Editors.Single (), propertyViewModel.Property);
+
+			Title = this.viewModel.PropertyDisplay;
+			this.ButtonDone.Title = Properties.Resources.CreatBindingTitle;
+
+			foreach (BindingSource item in this.viewModel.BindingSources.Value) {
```
ermau: This will throw an `NRE` if `BindingSources` hasn't loaded yet, it's an `AsyncValue<T>` which means `Value` will contain a default value until it's loaded. You should await the task on it before filling out this list.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,276 @@
...
+			this.PathBox.AddSubview (pathSelectorControl);
+
+			this.PathBox.AddConstraints (new[] {
+				NSLayoutConstraint.Create (pathSelectorControl, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this.PathBox, NSLayoutAttribute.Top, 1f, 5f),
+				NSLayoutConstraint.Create (pathSelectorControl, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this.PathBox, NSLayoutAttribute.Left, 1f, 5f),
+				NSLayoutConstraint.Create (pathSelectorControl, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this.PathBox, NSLayoutAttribute.Width, 1f, -10f),
+				NSLayoutConstraint.Create (pathSelectorControl, NSLayoutAttribute.Height, NSLayoutRelation.Equal, this.PathBox, NSLayoutAttribute.Height, 1f, -10f)
+			});
+
+			this.viewModel.PropertyChanged += (sender, e) => {
```
ermau: You're missing a check for `CanCreateBinding`, so the dialog allows you to create invalid bindings.

ermau: `PropertyDisplay` can also change.

vancura:
> It actually does have a header, you can see it in the second screenshot. It's dependent on the selected binding type.

Can we display it at all times?

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingTypeSelectorControl.cs
CartBlanche: Forgot to remove it. It was replaced by your TypeSelectorControl.

ermau:
> Can we display it at all times?

I don't know that it makes sense to show it at all times. What would we show in the header for the scenario in the current screenshot?

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingObjectSelectorControl.cs
```diff
@@ -0,0 +1,213 @@
...
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class BindingObjectSelectorControl : NSView
+	{
+		internal class ObjectOutlineView : NSOutlineView
+		{
+			private IReadOnlyList<ObjectTreeElement> viewModel;
+			public IReadOnlyList<ObjectTreeElement> ViewModel {
```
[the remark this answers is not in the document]
CartBlanche: Done!

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,276 @@
...
+	{
+		private readonly CreateBindingViewModel viewModel;
+		private readonly PropertyEditorSelector editorSelector = new PropertyEditorSelector ();
+
+		private const string BindingPropertiesIdentifier = "BindingProperties";
+		private const string FlagPropertiesIdentifier = "FlagProperties";
+
+		internal CreateBindingViewModel ViewModel => this.viewModel;
+
+		internal BindingEditorWindow (IHostResourceProvider hostResources, PropertyViewModel propertyViewModel)
```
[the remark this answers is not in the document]
CartBlanche: Done!

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,276 @@
...
+
+				this.BindingPropertiesView.AddSubview (nSView);
+
+				this.BindingPropertiesView.AddConstraints (new[] {
+					NSLayoutConstraint.Create (nSView, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this.BindingPropertiesView, NSLayoutAttribute.Top, 1f, controlTop),
+					NSLayoutConstraint.Create (nSView, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this.BindingPropertiesView, NSLayoutAttribute.Left, 1f, 16f),
+					NSLayoutConstraint.Create (nSView, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this.BindingPropertiesView, NSLayoutAttribute.Width, 1f, -20f),
+				});
+
+				controlTop += PropertyEditorControl.DefaultControlHeight;
```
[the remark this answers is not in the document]
CartBlanche: Done!

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingResourceSelectorControl.cs
```diff
@@ -0,0 +1,249 @@
...
+
+				return new NSObjectFacade (element);
+			}
+
+			public override bool ItemExpandable (NSOutlineView outlineView, NSObject item)
+			{
+				var target = (item as NSObjectFacade).Target;
+				switch (target) {
+					case IGrouping<ResourceSource, Resource> kvp:
+						return kvp.Count() > 0;
```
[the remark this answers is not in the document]
CartBlanche: Done.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingPathSelectorControl.cs
```diff
@@ -0,0 +1,339 @@
...
+		{
+			this.viewModel = viewModel;
+			this.targetName = targetName;
+		}
+
+		public override nint GetChildrenCount (NSOutlineView outlineView, NSObject item)
+		{
+			nint childCount;
+			if (item == null) {
+				childCount = this.viewModel != null ? this.viewModel.Count () + 1 : 0;
```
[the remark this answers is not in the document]
CartBlanche: Done

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingObjectSelectorControl.cs
```diff
@@ -0,0 +1,213 @@
...
+					throw new ArgumentNullException (nameof (viewModel));
+
+				ViewModel = viewModel;
+			}
+
+			public override nint GetChildrenCount (NSOutlineView outlineView, NSObject item)
+			{
+				var childCount = 0;
+				if (item == null) {
+					childCount = this.ViewModel != null ? this.ViewModel.Count () : 0;
```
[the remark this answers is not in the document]
CartBlanche: Done

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingObjectSelectorControl.cs
```diff
@@ -0,0 +1,213 @@
...
+				Initialize ();
+			}
+
+			[Export ("validateProposedFirstResponder:forEvent:")]
+			public bool ValidateProposedFirstResponder (NSResponder responder, NSEvent forEvent)
+			{
+				return true;
+			}
+
+			public void Initialize ()
```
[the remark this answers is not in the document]
CartBlanche: It shouldn't be.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingPathSelectorControl.cs
```diff
@@ -0,0 +1,339 @@
...
+			} else {
+				var target = (item as NSObjectFacade).Target;
+				switch (target) {
+					case PropertyTreeElement propertyTreeElement:
+						IReadOnlyCollection<PropertyTreeElement> propertyTrees = propertyTreeElement.Children.Task.Result;
+						childCount = propertyTrees.Count;
+						break;
+
+					case string targetName:
+						childCount = this.viewModel.Count ();
```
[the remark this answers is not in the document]
CartBlanche: Done.

vancura:
@ermau: Ok, fair enough :)

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingPathSelectorControl.cs
CartBlanche: This should have been small.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
CartBlanche: Sure.

CartBlanche:
https://github.com/xamarin/Xamarin.PropertyEditing/pull/561#discussion_r268811108
@ermau Not sure what you mean by "marked as resolved" and not fixed here?

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
CartBlanche: Thanks.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingObjectSelectorControl.cs
```diff
@@ -0,0 +1,213 @@
...
+					case TypeInfo info:
+						childCount = 0;
+						break;
+					default:
+						childCount = 0;
+						break;
+					}
+				}
+
+				return childCount;
```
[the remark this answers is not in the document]
CartBlanche: Yes. I used the variable breaks for easier debugging, but definitely not needed.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/CreateValueConverterWindow.cs
CartBlanche: Not sure how I lost this change. Will be fixed in my next push.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingEditorWindow.cs
```diff
@@ -0,0 +1,276 @@
...
+			this.ValueConverterPopup.Activated += (o, e) => {
+				if (this.ValueConverterPopup.Menu.HighlightedItem.RepresentedObject is NSObjectFacade facade) {
+					this.viewModel.SelectedValueConverter = (Resource)facade.Target;
+				}
+			};
+
+			RepopulateValueConverterPopup ();
+
+			this.AddConverterButton.Activated += (sender, e) => {
+				this.viewModel.SelectedValueConverter = CreateBindingViewModel.AddValueConverter;
```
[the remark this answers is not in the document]
CartBlanche: By command do you mean wrap it in a more discreet method call like `ViewModel.RequestAddValueConverter();`?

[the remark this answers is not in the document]
ermau: All other button invocations are done via `ICommand`

vancura:
Your redline 🐒 says hi.

## Basic Window Metrics

image

- [ ] Spacing around the window edge should be `20pt`, you have `29pt`.
- [ ] Vertical space between sections should be `23pt`, you have `28pt`.
- [ ] Please make the cancel button wider, specify minimum size to `80pt`.
- [ ] No help button? Just asking, if we don't have the contents, no worries. But help is always good.

## Header Section

image

- [ ] Use standard-sized dropdown (it seems you use tiny style).
- [ ] Title font size should be `13pt`, you seem to have `11pt`.
- [ ] Missing colon and capitalization: you have `Binding type` istead of `Binding Type:`.

## Table Section

image

- [ ] Path vs. checkbox header should use `20pt` space, but uses circa `60pt`.
- [ ] Please use standard line height in the tree view, lines are too separated.

## Bottom Section

image

- [ ] The title `Converter` is missing a colon (should be `Converter:`) and follow the same font size as specified above, in the header section (`13pt`).
- [ ] Space between dropdown and the plus button should be `10pt`, you have something around `136pt` (seems it's due to right-alignment of the plus button and automatic sizing of the dropdown; instead the dropdown needs to be fit the whole width.
- [ ] Contents of the bottom part should be vertically centered.

## Table Margins

image

- [ ] Please remove this padding, it shouldn't be there.

## Type Selector Window Metrics

image

- [ ] Spacing around the window edge should be `20pt`, you have `9pt`.

## Other Selector Window Inconsistencies

image

- [ ] Use standard filtering control.
- [ ] Align button set with the contents.

Your redline 🐒 thanks and says bye.

CartBlanche:
@vancura Here are some updated screenshots, based on using Figma to double check spacing and heights. These are 100% complete, but want to post these for you check before I push more pixels around.

Screen Shot 2019-04-11 at 16 45 50
Screen Shot 2019-04-11 at 16 46 11
Screen Shot 2019-04-11 at 16 46 42
Screen Shot 2019-04-11 at 16 47 23

CartBlanche:
@vancura 
> Please use standard line height in the tree view, lines are too separated.
We are using the default height for each tree node.

> No help button? Just asking, if we don't have the contents, no worries. But help is always good.

No help button for now. This may be added in future.

vancura:
This is looking amazing, well done @CartBlanche!

Screen Shot 2019-04-11 at 16 45 50

---

> I don't think the EditBox in the PathSelector looks good flush with box edge.

Can you try something like this? I removed the gap between the EditBox and the PathSelector, and added some space around the label inside the EditBox.

55967068-f142dd00-5c79-11e9-9038-47904faa766d

---

> I don't think we can make the type selector flush to the box, as it would clip the Show All Assemblies checkbox, on the left.

Can we have something like this?

Untitled-1

Changes:
- unified size of both EditBoxes in headers,
- removed the gap between EditBoxes and the selectors (see above),
- removed horizontal gap between the left selector and the edge of the table,
- moved the _Show All Assemblies_ slightly right to accommodate for lost space,

You said:

> We are using the default height for each tree node.

Well, in that case why both these selectors have different node height? From what I see here they contain a similar type of information, so they should have the same metrics. When next to each other it's odd how the node height differs.

---

55967071-f142dd00-5c79-11e9-98fa-b587e6af4192

A couple of changes:
- increased horizontal space between the right edge of the window and the right edge of the contents by `4pt`, so it's `20pt` as everywhere else,
- decreased vertical space between the top edge of the window and the top edge of the _Value converter name_ by `4pt`, so it's `20pt` as everywhere else,
- (I didn't do this as my SF fonts are broken now): please reword _Value converter name_ to _Value Converter Name:_ (capitalization and colon on end),
- moved buttons set, so the buttons are `20pt` from the edge of the window,
- changed the width of the buttons to `80pt`, make sure the width is always at least this size (of course if verbose languages don't fit, it scales, but `80pt` needs to be minimum),
- please, make sure the _Select_ button is default (blue) once a type is selected (so it's possible to press enter to confirm the selection),
- increased the vertical gap between button set and the _Show All Assemblies_ checkbox by `12pt`,
- optional: if it's possible to get the filter EditBox and the selector close to each other, similar to the screenshots above, this would be a killer.

CartBlanche:
@vancura I think these new changes look good....
Screen Shot 2019-04-12 at 18 57 47

As we are reusing the `Type Selector Control` in the `CreateValueConvertorWindow` the checkbox is still offset. Should we not keep the similar aesthetic? If not I could make further changes so it is flush, but wanted to check with you first before I do that.
Screen Shot 2019-04-12 at 18 58 14

Xamarin.PropertyEditing/ViewModels/CreateBindingViewModel.cs
```diff
@@ -674,5 +685,18 @@ IReadOnlyList<object> IProvidePath.GetItemParents (object item)
 				throw new ArgumentException();
 			}
 		}
+
+		private readonly RelayCommand requestAddValueConverterCommand;
+		public ICommand RequestAddValueConverter => this.requestAddValueConverterCommand;
+
+		private bool CanRequestAddValueConverter ()
+		{
+			return TargetPlatform.BindingProvider != null;
```
CartBlanche: @ermau Any other checks required to validate if the `AddValueConverter` button should be enabled when the BindingSource changes?

vancura:
This is awesome! Thanks for all the hard work @CartBlanche!

Xamarin.PropertyEditing.Tests/CreateBindingViewModelTests.cs
```diff
@@ -192,7 +193,12 @@ public async Task ValueConverters ()
 
 			await vm.ValueConverters.Task;
 			Assert.That (vm.ValueConverters.Value, Contains.Item (visi));
-			Assert.That (vm.ValueConverters.Value.Count, Is.EqualTo (3)); // visi, No Converter, Request Converter
+
+			if (vm.IncludeAddValueConverter) {
```
CartBlanche: @ermau we were already checking that the `ValueConverters.Value.Count` was correct in this test. As `GetValueConvertersAsync ()` is called and therefore `AddValueConverter` is added or not, via the `CreateBindingViewModel` constructor. Do we still need separate tests?

CartBlanche:
@ermau Yes It was a subscription issue and I wasn't setting the Appearance when creating the BindingDialog or the ValueConvertor Dialog. Both fixed now.

CartBlanche:
@ermau I'm not seeing this behaviour. The only thing that has been a theme bug for a while is the fact the wrong tab is on by default. This isn't related to the Binding Dialog change as it was around before then. I've pushed a fix for the tabs to be properly synced, as that needed fixing anyway.

CartBlanche:
@ermau could you please post a video of the behaviour you are seeing?

ermau:
I don't have a screen recorder on mac, but here are my exact steps:

1. Scroll down to the "String" property, create binding
2. Cancel create binding dialog
3. Switch to dark theme
4. Create binding on "String"

You will get the light themed dialog against dark:

Screen Shot 2019-05-29 at 13 02 18

Then when you close that dialog you'll get the dark themed one:
Screen Shot 2019-05-29 at 13 04 24

CartBlanche:
I see it now. Very odd. Investigating.

CartBlanche:
Right, events not being unsubscribed correctly. Appears to be fixed now.

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BaseSelectorOutlineView.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using AppKit;
+using Foundation;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public class BaseSelectorOutlineView : NSOutlineView
```
ermau:
```suggestion
	internal class BaseSelectorOutlineView : NSOutlineView
```

CartBlanche:
OK Test both scenarios this time. Think I have the right fix for it. Doing a little more testing.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -119,6 +135,9 @@ private void PopUpContextMenu ()
 				});
 			}
 
+			if (this.viewModel.SupportsBindings)
+				this.viewModel.CreateBindingRequested += OnBindingRequested;
```
ermau: This should be happening when the view model is assigned, not when the popup is opened. Because you assign this here and only unassign it when the binding is requested, each time you open the popup before actually hitting create binding, you're assuring that an additional binding dialog will appear.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -236,5 +255,19 @@ private void OnResourceRequested (object sender, EventArgs e)
...
+
+		private void OnBindingRequested (object sender, CreateBindingRequestedEventArgs e)
+		{
+			var bindingEditorWindow = new BindingEditorWindow (this.hostResources, this.viewModel) {
+				Appearance = EffectiveAppearance,
+			};
+
+			var result = (NSModalResponse)(int)NSApplication.SharedApplication.RunModalForWindow (bindingEditorWindow);
+			if (result == NSModalResponse.OK) {
+				e.BindingObject = bindingEditorWindow.ViewModel?.SelectedObjects?.Single ();
```
ermau: Neither the view model nor `SelectedObjects` should ever be null, so why are we treating them like they might be?

Xamarin.PropertyEditing.Mac/Controls/BindingEditor/BindingPathSelectorControl.cs
```diff
@@ -0,0 +1,173 @@
...
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class BindingPathSelectorControl 
+		: NotifyingView<CreateBindingViewModel>
+	{
+		private PathOutlineView pathOutlineView;
+		internal const string PathSelectorColumnColId = "PathSelectorColumn";
+
+        public NSTextField CustomPath { get; }
```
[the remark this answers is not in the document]
CartBlanche: Fair point.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -75,23 +75,25 @@ public void SetDefaultNextKeyView (NSView child)
...
-				view = TableView.GetView (0, row, makeIfNecessary: false);
-				ctrl = (view as EditorContainer)?.EditorView?.NativeView as PropertyEditorControl;
-			} while (row > 0 && ctrl == null);
-
-			if (ctrl != null) {
-				ctrl.LastKeyView.NextKeyView = FirstKeyView;
-				ctrl.UpdateKeyViews ();
-			} else if (row == 0 && view is PanelHeaderEditorControl header) {
-				header.SetNextKeyView (FirstKeyView);
+			if (TableView != null) {
```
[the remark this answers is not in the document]
CartBlanche: I think this is question of coding style and I prefer the other way.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
CartBlanche: Just being defensive.

CartBlanche:
@ermau I'm unable to find an elegant solution to the event subscription issue. Only found hacks, which I don't want to push as they weren't working 100% either. Can't see the woods for the trees anymore. Needs a new pair of eyes, when you have a sec.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
ermau: To me being defensive means things continue to work in an unforeseen circumstance. This won't do that, the error already occurred by these being null, and so it'll just hide the exception making it harder to find.

ermau:
@CartBlanche You needed to dig deeper. Your current implementation of the event subscription (in the `ViewModel`) is the correct one; the actual issue is separate.

We know that we're getting two dialogs, so we check for double subscription. We see one subscription when the property shows up, and then we see another when the theme changes. What we don't see is an unsubscribe at any point. Combine that with that one dialog is the old theme and we can deduce that when we change the theme, the property buttons are being re-created (as the old one with the old theme is still firing the event.) Now, that's an issue by itself, we should be reusing them when possible but since we may not always be able to, we need to ensure old discarded ones aren't still attached to anything by setting its View Model to `null`. A change was made for editors a while back to ensure this, but when property button moved around it was not included, I've comitted the fix to your branch.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
CartBlanche: Ok.

Issue #431, opened 2018-10-10
[Mac] Brush preview and text need to move up 1 or 2 pixels...


CartBlanche:
Please refer to screenshot below...

<img width="689" alt="screen shot 2018-10-10 at 13 37 20" src="https://user-images.githubusercontent.com/271363/46755092-3584fe80-cc92-11e8-8ea0-6971d13bafbb.png">

> VS bug [#703873](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/703873)

bleroy:
Fixed in current build: 
<img width="315" alt="image" src="https://user-images.githubusercontent.com/1165609/55647878-a74a8a80-5793-11e9-88bb-9d0b3d836457.png">

PR #563, opened 2019-03-15
Tweak the vertical margins in the color resources dialogs.

### LINKED ISSUE #432: [Mac] In Brush Resource selector, Colour preview is 1 pixel too high, so that highlight appear correctly.
Please refer to screenshot below...

screen shot 2018-10-10 at 13 37 53

> VS bug [#703872](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/703872)

bleroy:
Fixes #432, and probably also #431


Xamarin.PropertyEditing.Mac/Controls/Custom/CommonBrushLayer.cs
```diff
@@ -68,6 +70,7 @@ public static CALayer CreateBrushLayer (CommonBrush brush)
 		public override void LayoutSublayers ()
 		{
 			base.LayoutSublayers ();
+			Frame = new CGRect (Frame.X, Bounds.Y + VerticalMargin + 0.5, Bounds.Width, Bounds.Height - VerticalMargin * 2);
```
CartBlanche: Is it worth making the `0.5` and `2` constants?

CartBlanche: Also just a preference rather than anything, I would tend to wrap the `VerticalMargin * 2` in brackets just so it's very clear that is takes precedent. I know that as programmers we should all know BIDMAS, but at a glance IMO, it just makes things clearer.

ermau: Why the extra 0.5 on only one side?

bleroy: Because we're not adjusting the horizontal position at all here, just vertical (the color preview is vertically centered, but horizontally left-aligned), and vertical happens to be "naturally" (by which I mean if you let it do what it wants to do) one more pixel on one size than on the other. The 0.5 pixel restores parity between the margins above and below.

bleroy: 2, certainly not (it's just reflecting that a control has two vertical ends: up and down; it's not a measure of anything, and couldn't have another meaningful value). 0.5 is a one-time fudge that is also not going to be adjusted, reasonably, but a constant could make it more explicit what it's for.

PR #477, opened 2018-12-13
[Mac] Adds implementation to handle TimeSpan types

netonjm:
This PR adds the implementation to handle TimeSpan types.

timespan

PR #473, opened 2018-12-13
[Mac] Adds implementation to handle char types

netonjm:
Adds implementation to handle char types with a new Editor

PR #573, opened 2019-03-22
[mac] Refactor & Add TimeSpan / Char editors

ermau:
This PR:
 - Refactors common entry elements into a common base class
 - Changes string commit behavior to blur/enter like the other editors
 - Adds a TimeSpan editor based on new entry base (replacing #477)
 - Adds a char editor base on new entry base (replacing #473)

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/CharEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/CharEditorControl.cs
@@ -1,0 +1,50 @@
+using System;
+using Xamarin.PropertyEditing.Mac.Resources;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class CharEditorControl
+		: EntryPropertyEditor<char>
+	{
+		public CharEditorControl (IHostResourceProvider hostResources)
+			: base (hostResources)
+		{
+		}
+
+		protected override EntryPropertyEditorDelegate<char> CreateDelegate (PropertyViewModel<char> viewModel)
+		{
+			return new CharDelegate (viewModel);
+		}
+
+		protected override void UpdateAccessibilityValues ()
+		{
+			base.UpdateAccessibilityValues ();
+			Entry.AccessibilityTitle = string.Format (LocalizationResources.AccessibilityChar, ViewModel.Property.Name);
+		}
+
+		protected override string GetValue (char value)
+		{
+			return (value == default (char)) ? null : value.ToString();
+		}
+
+		private class CharDelegate
+			: EntryPropertyEditorDelegate<char>
+		{
+			public CharDelegate (PropertyViewModel<char> viewModel)
+				: base (viewModel)
+			{
+			}
+
+			protected override char GetValue (string value)
+			{
+				return (String.IsNullOrEmpty (value) ? default (char) : value[0]);
+			}
+
+			protected override bool CanGetValue (string value)
+			{
+				return Char.TryParse (value, out _);
+			}
+		}
+	}
+}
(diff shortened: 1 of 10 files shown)


alanmcgovern:
The original build had a random failure, we should look into this: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2521105

Xamarin.PropertyEditing.Mac/Controls/CharEditorControl.cs
```diff
@@ -0,0 +1,50 @@
...
+
+		protected override EntryPropertyEditorDelegate<char> CreateDelegate (PropertyViewModel<char> viewModel)
+		{
+			return new CharDelegate (viewModel);
+		}
+
+		protected override void UpdateAccessibilityValues ()
+		{
+			base.UpdateAccessibilityValues ();
+			Entry.AccessibilityTitle = string.Format (LocalizationResources.AccessibilityTimeSpan, ViewModel.Property.Name);
```
CartBlanche: I suspect this might be a copy/paste error, but the `Entry.AccessibilityTitle` should probably be assigned `LocalizationResources.AccessibilityChar`?

Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
```diff
@@ -11,11 +11,13 @@ public class MockSampleControl : MockControl
 	{
 		public MockSampleControl ()
 		{
+			AddProperty<TimeSpan> ("TimeSpan", ReadWrite, valueSources: ValueSources.Local | ValueSources.Resource | ValueSources.Binding);
```
CartBlanche: Not sure `TimeSpan` should support `ValueSources.Resource`  here...
Screen Shot 2019-03-25 at 11 52 12

ermau: We might not have any example resources in the standalone, but there's nothing wrong with this.

ermau:
> Readonly versions of Char and TimeSpan do not exist in the property list, so we don't know how they will look/behave.

I tested them before pushing, but we can add permanent versions.

CartBlanche:
@ermau one thing I was wondering, should the TimeSpan support tabbing between the 3 time elements?

CartBlanche:
> I tested them before pushing, but we can add permanent versions.
I mentioned it because in the past sometimes the ReadOnly versions show up other issues. Might not be the case here.

Xamarin.PropertyEditing.Tests/MockControls/MockSampleControl.cs
netonjm: makes sense to include here a NSDatePicker with only time elements (like we have with Dates)

 this.datePicker = new NSDatePicker {
				ControlSize = NSControlSize.Small,
				DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond,
				DatePickerStyle = NSDatePickerStyle.TextFieldAndStepper,
				Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Small)),
				TranslatesAutoresizingMaskIntoConstraints = false
			};

netonjm:
makes sense to include here a NSDatePicker with only time elements (like we have with Dates)

this.datePicker = new NSDatePicker {
ControlSize = NSControlSize.Small,
DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond,
DatePickerStyle = NSDatePickerStyle.TextFieldAndStepper,
Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Small)),
TranslatesAutoresizingMaskIntoConstraints = false
};

CartBlanche:
@netonjm that will only be useful if we can set the Min/Max values for that component. Can we? Looking at the TimeSpan this morning, I don't think it the native DateTime control would handle it. Hence my suggestion for more of a custom solution. Or maybe that can be done in a future update?

ermau:
Guys this is the original solution submitted just cleaned up, if we want a fancier solution we can tackle that later.

netonjm:
Makes sense Eric

PR #575, opened 2019-03-25
[build] Strongname this so it can be referenced by other strongnamed …

alanmcgovern:
…libs


netonjm:
Windows seems not happy

netonjm:
mmm we need change the manifest to point the correct key hash for internals

alanmcgovern:
Yup - the build failure gave the info i needed to update the `InternalsVisibleTo` :)

PR #577, opened 2019-03-25
[mac] Expose header control

### LINKED ISSUE #485: Hide Header section (Name/type fields) in API
In MD there are no Name and Type fields, it would be great a hidden flag for this.

https://user-images.githubusercontent.com/1587480/50037383-d7b5bf80-0010-11e9-83df-0c171739e7ff.png

ermau:
Fixes #485

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
@@ -41,6 +41,12 @@ namespace Xamarin.PropertyEditing.Mac
 			Initialize ();
 		}
 
+		public bool ShowHeader
+		{
+			get => this.propertyList.ShowHeader;
+			set => this.propertyList.ShowHeader = value;
+		}
+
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
```diff
@@ -41,6 +41,12 @@ public PropertyEditorPanel (NSCoder coder) : base (coder)
 			Initialize ();
 		}
 
+		public bool ShowHeader
+		{
+			get => this.propertyList.ShowHeader;
+			set => this.propertyList.ShowHeader = value;
+		}
+
```
alanmcgovern:
```
		public bool ShowHeader
		{
			get { return this.showHeader; }
			set
			{
				if (this.showHeader == value)
					return;

				this.showHeader = value;
				if (this.dataSource != null) {
					this.dataSource.ShowHeader = false;
					this.propertyTable.ReloadData ();
				}
			}
```
Did we mean to hardcode false?

Issue #510, opened 2019-01-30
[Mac][Accessibility] Press Tab in SearchTextField doesn't focus the first element of the PropertyPanel


netonjm:
![badfocusorder](https://user-images.githubusercontent.com/1587480/52006239-f4785a80-24cb-11e9-9065-685ccfec0fa6.gif)

netonjm:
Maybe is needed a Window.RecalculateKeyViewLoop after generate all property views

CartBlanche:
I tried calling `Window.RecalculateKeyViewLoop` or  setting `NextKeyView`, but neither works. May need some help it getting this one sorted.

netonjm:
Did you check if could be related override PreviousResponder and NextResponder? Maybe the first element has no PreviousResponder or something is breaking the current Focus chain?

CartBlanche:
@netonjm I'm back to looking at this, so will double check. I think I checked this last time and everything looked fine to me.

PR #578, opened 2019-03-26
[Mac] Handle tabbing into the NSOutlineView

alanmcgovern:
Fixes - https://github.com/xamarin/Xamarin.PropertyEditing/issues/510
If the outline view is made the first responder we should select
the first row (if there's no selection) and then we should make
it the first responder.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyList.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyList.cs
@@ -97,7 +97,7 @@ namespace Xamarin.PropertyEditing.Mac
 			this.propertyTable.BackgroundColor = this.hostResources.GetNamedColor (NamedResources.PadBackgroundColor);
 		}
 
-		public void UpdateExpansions()
+		public void UpdateExpansions ()
 		{
 			((PropertyTableDelegate)this.propertyTable.Delegate).UpdateExpansions (this.propertyTable);
 		}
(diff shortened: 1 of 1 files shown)


alanmcgovern:
@CartBlanche this only helps when tabbing forward

CartBlanche:
@alanmcgovern It looks to me like when trying to reverse tab, back to the `propertySearch` that the `BecomeFirstResponder` code it reseting the focus.
I'm will also investigate (after the Binding changes) a ghost tab, if the `propertyList` is empty.
Definitely MUCH better than what we currently have.

CartBlanche:
Current Tabbing. Single out of PropertyFilter.
[https://youtu.be/gM-YjtvgIFo](https://youtu.be/gM-YjtvgIFo)

CartBlanche:
@ermau, @alanmcgovern Please either take a look at the code or watch the video link above.
If you tab forwards all the way to the end, it single tabs correctly all the way through the `PropertyList` and the `PropertyFilter` and round again.
The *only* glitch is having to double tab backwards from the very 1st control of the `PropertyList` back to the `PropertyFilter` control. So we're 98% of the way there. It is about 1000% better than what we currently have, as tabbing right now doesn't work at all.

CartBlanche:
Add a new issue referring to the double back tab from the PropertyPanel - https://github.com/xamarin/Xamarin.PropertyEditing/issues/589

PR #579, opened 2019-03-27
Fixes NRE trying to set a null NSColor when no host resource name is set

netonjm:


alanmcgovern:
This is not the correct way to fix this.

alanmcgovern:
this is what i shared earlier https://github.com/xamarin/designer/pull/1792#discussion_r267446592

PR #580, opened 2019-03-29
Windows Variants

ermau:
Few things still need resolving from design (@vancura):
 - [x] VS color service colors

I've deviated from the design to adjust the button colors to match those of the property button. We will need service colors to attach the variation blocks background and foreground colors to.

 - [x] Variation font size

For the font size, there's only really one environment font setting for VS. We can either try to do some kind of % font size, bump it up to full size, or try to find another font service font to attach it to.

**Screenshots:**
Xamarin PropertyEditing Windows Standalone_2019-03-29_15-12-30
Xamarin PropertyEditing Windows Standalone_2019-03-29_15-03-05
Xamarin PropertyEditing Windows Standalone_2019-03-29_15-12-45
Xamarin PropertyEditing Windows Standalone_2019-03-29_15-12-42


vancura:
@ermau This is looking so good! Thank you! One thing, could you please grab a screenshot how it looks when there are more variants than what could fit in one line? Does it overflow or wrap?

vancura:
And one more, could you please grab a HiDPI screenshot? I'd like to see how the line aligns in a higher resolution. Thanks!

ermau:
@vancura It currently overflows, I can probably make it wrap without too much difficulty. I don't think I have a way to get you a high dpi screenshot from Windows.

vancura:
I'll try to compile and will get back to you.

vancura:
This is great, thanks @ermau!

image

Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
```diff
@@ -64,6 +161,54 @@ private void OnLoaded (object sender, RoutedEventArgs e)
...
+		{
+			var variation = CreateVariantWindow.RequestVariant (this, this.pvm.Property);
+			if (variation != null)
+				e.Variation = Task.FromResult (variation);
+		}
+
+		private void OnRemoveVariantClicked (object sender, RoutedEventArgs e)
+		{
+			var parent = (FrameworkElement) VisualTreeHelper.GetParent (this);
+			while (!(parent != null && parent.DataContext is PanelGroupViewModel)) {
```
bleroy: I think we have a helper in `XamlHelper` that looks up the parent chain for a specific type already.

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
```diff
@@ -341,6 +349,30 @@ protected override void OnClearProperties ()
...
+
+			var list = group.GetList (viewModel);
+
+			int index = list.IndexOf (viewModel);
+			if (index == -1)
+				throw new KeyNotFoundException ($"{nameof (viewModel)} was not found");
+
+			if (++index == list.Count)
+				return true;
+			if (list[index] is PropertyViewModel pvm) {
```
bleroy: that's just `return (++index == list.Count) || (list[index] is PropertyViewModel pvm && !Equals (viewModel.Property, pvm.Property));`

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -226,20 +240,45 @@ private void OnErrorsChanged (DataErrorsChangedEventArgs e)
...
-				var properties = this.editors
-					.OfType<PropertyViewModel> ()
-					.Where (evm => Equals (evm.Property, pvm.Property) && evm.Variation != null)
-					.ToDictionary (evm => evm.Variation);
+				var variationsTask = GetVariationsAsync (property);
+
+				PropertyViewModel baseVm = null;
+				var properties = new Dictionary<PropertyVariation, PropertyViewModel> ();
+				foreach (PropertyViewModel pvm in this.editors.OfType<PropertyViewModel> ()) {
+					if (!Equals (property, pvm.Property))
```
bleroy: You can move that to a `Where` in the `foreach`.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -532,34 +571,27 @@ private async void OnObjectEditorPropertiesChanged (object sender, NotifyCollect
...
 			if (baseVm.HasVariations) {
 				using (await AsyncWork.RequestAsyncWork (this)) {
 					var variants = await GetVariationsAsync (property);
-					for (int i = 0; i < variants.Length; i++) {
-						foreach (PropertyVariation variant in variants[i]) {
-							if (skipped != null && skipped.Contains (variant))
-								continue;
-
-							vms.Add (GetViewModel (property, variant));
+					if (baseVm.HasVariantChildren = variants.Count > 0) {
```
bleroy: I'd move that assignment out of the test.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -532,34 +571,27 @@ private async void OnObjectEditorPropertiesChanged (object sender, NotifyCollect
...
+						foreach (PropertyVariation variant in variants) {
```
bleroy: I think that's `vms.AddRange (variants.Select (v => CreateViewModel (property, v)))`.

Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
```diff
@@ -56,6 +64,95 @@ public bool ShowPropertyButton
...
+
+		public override void OnApplyTemplate ()
+		{
+			this.addButton = GetTemplateChild ("addVariant") as ButtonBase;
+			this.removeButton = GetTemplateChild ("removeVariant") as ButtonBase;
+			if (this.removeButton != null) {
+				this.removeButton.Click += OnRemoveVariantClicked;
+			}
+
+			this.variationRow = (RowDefinition) GetTemplateChild ("variationRow");
```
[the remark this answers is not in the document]
ermau: The other ones may not be realized when there are not variant features, but the variant row will always be present.

Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
ermau: I'll swap, I remembered about the XamlHelpers after I wrote a bunch of this; didn't go back and check.

ermau: It finds a specific view parent type, this is looking for a specific view model type.

Xamarin.PropertyEditing/ViewModels/PanelViewModel.cs
ermau: I find this easier to read, no need to change.

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
ermau: True, but I don't find it that much cleaner for its high cost.

ermau:
> Reviewed the code, and tested the branch on Windows. Everything looks fine, except an invalid cast exception causing the app to crash when deploying the brush category.

Fixed.

vancura:
@ermau:

> For the font size, there's only really one environment font setting for VS. We can either try to do some kind of % font size, bump it up to full size, or try to find another font service font to attach it to.

How much is it now? Are we talking about the font size in these bubbles?

image

We need to tune the size finely, the font size is too low – on HiDPI it's okay, but the ClearType antialiasing does more harm than good here.

---

> VS color service colors

Variation bubbles:
- Background: `EnvironmentAccessKeyToolTip`
- Text: `EnvironmentAccessKeyToolTipText`

Do you also need the background colors for the overrides with variations?

vancura:
> It's currently 9px. I think Figma had it at 9 or 9.5. Yes, that font size.

What's the normal font size in this area? 100% scaling, I mean.

ermau:
> What's the normal font size in this area? 100% scaling, I mean.

Not sure what area you mean? The normal font size for all the other text? The default is 12px (in dip), but it's dynamic. It's controlled by the user in their environment settings, that's why I was suggesting we need a % of or something.

vancura:
Yup, but what's the percentage of the font usual font size used in the variation bubbles? What I mean: the ClearType antialiasing messes up readability on standard DPI, we need to increase font size slightly. How much I don't know, that's why I am asking.

ermau:
9/12 = 75%

vancura:
Could you please try `10pt` font, circa 83%?

ermau:
@vancura 
Here's 83% rounded at original size of 12, 16, and 10 respectively:

Xamarin PropertyEditing Windows Standalone_2019-05-03_15-25-53
Xamarin PropertyEditing Windows Standalone_2019-05-03_15-26-05
Xamarin PropertyEditing Windows Standalone_2019-05-03_15-26-19

vancura:
This looks great to me!

Xamarin.PropertyEditing/ViewModels/PropertiesViewModel.cs
```diff
@@ -181,8 +182,13 @@ protected virtual async void OnSelectedObjectsChanged (object sender, NotifyColl
...
 					for (int i = 0; i < removedEditors.Length; i++) {
-						removedEditors[i] = this.objEditors[i];
-						INotifyCollectionChanged notifier = removedEditors[i]?.Properties as INotifyCollectionChanged;
+						IObjectEditor editor = this.objEditors[i];
+						if (editor == null)
+							continue;
+
+						removedEditors[i] = editor;
+						editor.PropertyChanged -= OnObjectEditorPropertyChanged;
+						INotifyCollectionChanged notifier = editor.Properties as INotifyCollectionChanged;
```
alanmcgovern: 👀

Issue #583, opened 2019-04-08
Rewrite Mac accessibility help text


ermau:
Most of the Mac accessibility text isn't all that useful, it often uses our name for the editor which isn't useful, and then the built in support lists the real name of the editor sometimes making it a repeat.

PR #584, opened 2019-04-08
[Mac] Centralise string resources.

### LINKED ISSUE #582: Move Mac string resources into central list
There's a .resx in Xamarin.PropertyEditing and one in Xamarin.PropertyEditing.Mac. We should move the resources from .Mac into the central one to reduce the number of translation assemblies and ensure we don't get duplicates.

CartBlanche:
Fixes #582


Xamarin.PropertyEditing/Xamarin.PropertyEditing.csproj
```diff
@@ -15,4 +15,9 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <ItemGroup>
```
ermau: Do we need this, given the above? cc @alanmcgovern

CartBlanche: @ermau That was auto added by the IDE.

alanmcgovern: This should be fine. It's just ensuring the autogenerated file shows up as a child of the 'resx' file in the solution pane. If we don't add it the odds are the IDE will continually try to re-add it.

Xamarin.PropertyEditing/Properties/Resources.resx
```diff
@@ -577,4 +578,172 @@
...
+    <value>Choose a {0}</value>
+    <comment>Choose a File or Directory</comment>
+  </data>
+  <data name="BrowseFileOrDirectory" xml:space="preserve">
+    <value>Browse for a {0}</value>
+    <comment>Browse for a File or Directory</comment>
+  </data>
+  <data name="RevealFileOrDirectory" xml:space="preserve">
+    <value>Reveal {0}</value>
+    <comment>Reveal a File or Directory</comment>
```
alanmcgovern: I'm just reviewing the strings now and i see a problem - we can't use string concatenation to produce localized values.

If we need the strings `Reveal a File` and also `Reveal a Directory` then to properly translate those we need to have two different strings in our resx and then choose the appropriate one in code. What we *cannot* do is the equivalent of:

```
myLabel.Text = string.Format (Properties.Resources. RevealFileOrDirectory, IsDirectory ? Properties.Resources.Directory : Properties.Resources.File)
```
This will not produce a valid string in languages other than english. We should review in a separate PR to ensure we use whole strings instead of concatenation :)

Xamarin.PropertyEditing/Properties/Resources.resx
```diff
@@ -577,4 +578,172 @@
...
+    <comment>Editor for Y Value</comment>
+  </data>
+  <data name="AccessibilityWidthEditor" xml:space="preserve">
+    <value>{0} Width Editor</value>
+    <comment>Editor for Width Value</comment>
+  </data>
+  <data name="AccessibilityHeightEditor" xml:space="preserve">
+    <value>{0} Height Editor</value>
+    <comment>Editor for Height Value</comment>
+  </data>
```
alanmcgovern: The comment's here should reflect how the string in the `value` element should be populated. We can give a few examples if needed.

In this case does it populate as `Margin Height Editor`? `Editor for Height Value` is a bit weird as a comment as it does not relate to the real string.

CartBlanche:
I'd think we'll handle the string formatting and better comments as part of  addressing this issue - https://github.com/xamarin/Xamarin.PropertyEditing/issues/583

PR #585, opened 2019-04-09
Convert from xammac library proj

alanmcgovern:


alanmcgovern:
This should be rebased/mergeg after the resx movement PR merges. There'll be fewer conflicts then!

CartBlanche:
@alanmcgovern I rebased and pushed, and on the build server everything appears to be fine, but locally tests are failing. Are you seeing that behaviour locally?

CartBlanche:
Looks like it may be VSM related.

PR #586, opened 2019-04-10
[Mac] Don't show 'Reset' for read-only properties

alanmcgovern:
This avoids PropertyPadObjectEditor throwing an exception in
SetValueAsync<T> as the property is not writeable.


alanmcgovern:
Should the property button be hidden if/when there are no items in the context menu? Once this, or a version of this, goes into master then readonly properties in VSMac when selecting a 'cs' file in the property pad will have no entries left.

alanmcgovern:
screencast: http://recordit.co/GVGCUT1NPu

ermau:
We should be disabling it, not hiding it.

PR #590, opened 2019-04-29
[Mac] Move appearance setting calls outside ViewDidChangeEffectiveApp…

### LINKED ISSUE #587: Move ViewDidChangeEffectiveAppearance calls
We need to move our effective appearance calls to a second method that we can call from the ctor and an override.

CartBlanche:
…erance, so it works on earlier MacOS versions.

Fixes #587 .

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/ColorEditorView.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/ColorEditorView.cs
@@ -43,6 +43,17 @@ namespace Xamarin.PropertyEditing.Mac
 		public virtual void OnPropertyChanged (object sender, PropertyChangedEventArgs e)
 		{
 		}
+
+		protected virtual void AppearanceChanged ()
+		{
+		}
+
+		public sealed override void ViewDidChangeEffectiveAppearance ()
+		{
+			base.ViewDidChangeEffectiveAppearance ();
+
+			AppearanceChanged ();
+		}
 	}
 
 	internal abstract class ColorEditorView : NotifyingView<SolidBrushViewModel>
--- a/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
@@ -197,16 +197,23 @@ namespace Xamarin.PropertyEditing.Mac
 				NSLayoutConstraint.Create (this.propertyList, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this, NSLayoutAttribute.Width, 1, 0),
 			});
 
-			ViewDidChangeEffectiveAppearance ();
+			UpdateResourceProvider ();
 		}
 
-		private void UpdateResourceProvider()
+		public override void ViewDidChangeEffectiveAppearance ()
 		{
-			this.propertyList.HostResourceProvider = HostResourceProvider;
-			this.header.HostResourceProvider = HostResourceProvider;
-			this.border.HostResourceProvider = HostResourceProvider;
+			base.ViewDidChangeEffectiveAppearance ();
 
-			ViewDidChangeEffectiveAppearance ();
+			UpdateResourceProvider ();
+		}
+
+		private void UpdateResourceProvider ()
+		{
+			if (this.propertyList != null) {
+				this.propertyList.HostResourceProvider = HostResourceProvider;
+				this.header.HostResourceProvider = HostResourceProvider;
+				this.border.HostResourceProvider = HostResourceProvider;
+			}
 		}
 
 		private void OnArrangeModeChanged (object sender, EventArgs e)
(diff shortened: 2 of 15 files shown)


CartBlanche:
Seems this VS4Mac PR https://github.com/xamarin/monodevelop/pull/552 may need this change before their PR can go through.

Xamarin.PropertyEditing.Mac/Controls/Custom/ColorEditorView.cs
```diff
@@ -43,6 +43,10 @@ public virtual void OnViewModelChanged (TViewModel oldModel)
 		public virtual void OnPropertyChanged (object sender, PropertyChangedEventArgs e)
 		{
 		}
+
+		protected virtual void AppearanceChanged ()
```
alanmcgovern: Can we add this declaration here:

```
public sealed override void ViewDidChangeChangeEffectiveAppearance ()
{
    base.ViewDidChangeEffectiveAppearance ();
    AppearanceChanged ();
}
```
This gives us two main advantage:
1) We are guaranteed that ViewDidChangeEffectiveAppearance is overridden once for all subclasses, so AppearanceChanged will only be invoked once every time ViewDidChangeEffectiveAppearance is invoked. Without this, if there are multiple subclasses then each of them may override ViewDidChangeEffectiveAppearance and invoke AppearanceChanged, which means we'll get multiple invocations of AppearanceChanged.
2) It ensures that no-one accidentally puts code in ViewDidChangeEffectiveAppearance in other subclasses.

On a related note, we should probably seal every override of this method for the reasons :)

Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
```diff
@@ -197,16 +197,21 @@ private void Initialize ()
...
-			ViewDidChangeEffectiveAppearance ();
+			UpdateResourceProvider ();
 		}
 
-		private void UpdateResourceProvider()
+		public override void ViewDidChangeEffectiveAppearance ()
+		{
+			base.ViewDidChangeEffectiveAppearance ();
+
+			UpdateResourceProvider ();
```
alanmcgovern: Should this method be renamed `AppearanceChanged` for consistency with every other implementation?

Xamarin.PropertyEditing.Mac/PropertyList.cs
```diff
@@ -91,10 +91,9 @@ public IHostResourceProvider HostResourceProvider
 
 		public override void ViewDidChangeEffectiveAppearance ()
 		{
-			if (this.propertyTable == null)
-				return;
+			base.ViewDidChangeEffectiveAppearance ();
 
-			this.propertyTable.BackgroundColor = this.hostResources.GetNamedColor (NamedResources.PadBackgroundColor);
+			UpdateResourceProvider ();
```
alanmcgovern: Should this method be renamed `AppearanceChanged` for consistency with every other implementation?

Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
CartBlanche: @alanmcgovern I did consider it, but currently it specifically only updates the respective ResourceProvider and not any other Appearance values for specific control. If you really want it it changed, I can do that.

Xamarin.PropertyEditing.Mac/Controls/Custom/ColorEditorView.cs
CartBlanche: @alanmcgovern When you say *here*, I assume you actually mean in the actual base class. Just asking to be 100% clear.

CartBlanche: Hmm Nm, This makes sense, but highlights I need to do this in the *other* base class.

Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
alanmcgovern: I suppose i'm just looking for a consistent set of `sealed override ViewDidChangeEffectiveAppearance` and `void UpdateAppearance` methods.

Let's put in a pin in this for now as i just did a quick code review of the rest of the codebase and have some concerns. I'll file an issue with my thoughts!

Xamarin.PropertyEditing.Mac/Controls/Custom/ColorEditorView.cs
CartBlanche: Done!

alanmcgovern: Ah yeah, i meant in this base class and all other places where the method is overridden :)

alanmcgovern:
🎉

PR #594, opened 2019-05-08
Changed Accessibility text to sound more natural when read out. Added…

CartBlanche:
Fixes https://github.com/xamarin/Xamarin.PropertyEditing/issues/583


Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -36,7 +36,7 @@ public PropertyButton (IHostResourceProvider hostResources)
 
 			this.hostResources = hostResources;
 
-			AccessibilityTitle = Properties.Resources.AccessibilityPropertiesButton;
+			AccessibilityTitle = string.Format (Properties.Resources.AccessibilityPropertiesButton, ViewModel.Property.Name);
```
ermau: `ViewModel` can't be set in the ctor, so this will always throw an `NRE`. Additionally, it wouldn't update when the button is re-used with a different property.

CartBlanche: Indeed!

CartBlanche:
@ermau You'll have to take a look and (VoiceOver) listen from here and tweak as appropriate.

PR #596, opened 2019-05-08
[Mac] Setting the Formatter by default breaks the RatioEditor rendering.

### LINKED ISSUE #553: [Mac] Ratio editor loses value when clicking the up button
In the standalone mac application, select all the text in the ratio property, then hit the up button. This should update the value, but clears it instead.

CartBlanche:
The simplest solution would have been to set Formatter property to null, but that blows up :/

So Marius has raised an issue for that on the Mac side here:
https://github.com/xamarin/xamarin-macios/issues/6032

and suggested this workaround
https://gist.github.com/Therzok/f735948ec766d2dca16af958913eef36

Which appears to work.

Fixes #553

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Extensions.cs
+++ b/Xamarin.PropertyEditing.Mac/Extensions.cs
@@ -1,0 +1,24 @@
+using System;
+using System.Runtime.InteropServices;
+using AppKit;
+using Foundation;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal static class Extensions
+	{
+		internal const string LIBOBJC_DYLIB = "/usr/lib/libobjc.dylib";
+
+		[DllImport (LIBOBJC_DYLIB, EntryPoint = "objc_msgSend")]
+		public extern static void void_objc_msgSend_intptr (IntPtr receiver, IntPtr selector, IntPtr arg1);
+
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing.Mac/Extensions.cs
```diff
@@ -0,0 +1,24 @@
...
+	{
+		internal const string LIBOBJC_DYLIB = "/usr/lib/libobjc.dylib";
+
+		[DllImport (LIBOBJC_DYLIB, EntryPoint = "objc_msgSend")]
+		public extern static void void_objc_msgSend_intptr (IntPtr receiver, IntPtr selector, IntPtr arg1);
+
+		const string selSetFormatter = "setFormatter:";
+		static readonly IntPtr selSetFormatter_Handle = ObjCRuntime.Selector.GetHandle (selSetFormatter);
+
+		internal static void SetFormatter (this NumericSpinEditor control, NSFormatter formatter)
```
Therzok:
```suggestion
		internal static void SetFormatter (this NSControl control, NSFormatter formatter)
```

Some other code might benefit from it.

Therzok: Ah, I see, it won't apply as suggested, as we're using NumericEditor.

PR #600, opened 2019-05-16
[Mac] Initial Variations implementation

CartBlanche:


CartBlanche:
Current L&F of 
Variations Dialog:
Screen Shot 2019-07-03 at 10 36 10

Variations Tree Rendering:
Screen Shot 2019-07-03 at 10 35 20

CartBlanche:
@vancura Could you please case your eye over the screenshots above ^^

vancura:
Hey,

- [ ] There's a huge space between dropdowns and buttons on the top image
- [ ] Wrong font size in the variation label (1pt lower, please)
- [ ] Please don't use "X" and "+" characters on add/remove. We have bitmaps for these.

ermau:
@vancura I'm curious about the change from the popup to a dialog, what was the motivation?

CartBlanche:
@vancura I don't that the bitmaps you speak of. 

I have and add and remove, but when used look like this, not like the figma...

Screen Shot 2019-07-03 at 17 31 02

CartBlanche:
@vancura 
Screen Shot 2019-07-04 at 18 16 10

Also please see my comment above about the missing images.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -24,6 +28,127 @@ public EditorViewModel ViewModel
...
+						var addVariantButton = new CommandButton {
+							Bordered = false,
+							Command = propertyViewModel.RequestCreateVariantCommand,
+							// TODO Awaiting Image from Vaclav Image = HostResources.GetNamedImage ("pe-list-add-16"),
+							Title = "+",
+							ToolTip = Properties.Resources.AddVariant,
+							TranslatesAutoresizingMaskIntoConstraints = false,
+						};
+
+						AddSubview (addVariantButton);
```
ermau:
```suggestion
						LeftEdgeView = addVariantButton;
```

This is what `LeftEdgeView` is meant for. When we add error handling, these two will draw over each other otherwise.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -24,6 +28,127 @@ public EditorViewModel ViewModel
...
+							endLineShape.StrokeColor = HostResources.GetNamedColor (NamedResources.ForegroundColor).CGColor;
+							Layer.AddSublayer (endLineShape);
+
+							this.variationLayersList.Add (endLineShape);
+						} else {
+							var midLineShape = new CAShapeLayer ();
+							var midLinePath = new NSBezierPath ();
+
+							midLinePath.MoveTo (new CGPoint (10, 0));
+							midLinePath.Append (new CGPoint[] { new CGPoint (10, 55) });
```
ermau: Do we need to duplicate all this when these are the only two different lines?

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -24,6 +28,127 @@ public EditorViewModel ViewModel
...
 					PropertyButton.ViewModel = value as PropertyViewModel;
+
+				UpdateVariations (value as PropertyViewModel);
+			}
+		}
+
+		private void UpdateVariations (PropertyViewModel propertyViewModel)
+		{
+			if (propertyViewModel != null && propertyViewModel.Property.CanWrite) {
+				if (propertyViewModel.HasVariations) {
```
ermau: This should be in the conditional above it. Otherwise if the property is writable but doesn't have variations, nothing will get cleaned up.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -30,6 +30,7 @@ public IHostResourceProvider HostResources
 
 		public const int DefaultControlHeight = 24;
 		public const int DefaultFontSize = 11;
+		public const int DefaultMenuFontSize = DefaultFontSize + 1;
```
ermau: We're already trying to get away from a "DefaultFontSize" by using the mac named sizes, let's not compound the problem.

Xamarin.PropertyEditing.Mac/Controls/PropertyContainer.cs
```diff
@@ -89,5 +91,7 @@ protected INativeContainer NativeContainer
 			},
 			TranslatesAutoresizingMaskIntoConstraints = false,
 		};
+
+		public IHostResourceProvider HostResources { get; private set; }
```
ermau: This could probably be `protected`

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantView.cs
```diff
@@ -0,0 +1,121 @@
...
+		: NSView
+	{
+		private const int ControlSpacing = 7;
+		private const int HorizontalControlSpacing = 10;
+		internal const int RightEdgeMargin = 12;
+
+		private IHostResourceProvider hostResources;
+
+		public CreateVariantView (IHostResourceProvider hostResources, CGSize windowWidthHeight, CreateVariantViewModel createVariantViewModel)
+		{
```
ermau: `ArgumentNullException` check for `createVariantViewModel`

ermau:
@vancura Should we have some sort of rollover like the property button to indicate the +/x are clickable?

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantView.cs
CartBlanche: Yes, missed it.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
CartBlanche: True.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
CartBlanche: Yes I should have refactored that.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
CartBlanche: OK

Xamarin.PropertyEditing.Mac/Controls/EntryPropertyEditor.cs
```diff
@@ -21,10 +21,9 @@ public EntryPropertyEditor (IHostResourceProvider hostResources)
 
 			RightEdgeConstraint = NSLayoutConstraint.Create (Entry, NSLayoutAttribute.Right, NSLayoutRelation.Equal, this, NSLayoutAttribute.Right, 1f, 0);
 			AddConstraints (new[] {
-				NSLayoutConstraint.Create (Entry, NSLayoutAttribute.CenterY, NSLayoutRelation.Equal, this,  NSLayoutAttribute.CenterY, 1f, 0),
```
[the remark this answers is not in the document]
CartBlanche: Because if it is `CentreY` constrained, when the control height changes it will position itself nicely in the horizontal centre. Like this...
Screen Shot 2019-07-15 at 18 29 18

The current `Bottom` constrain looks like this..
Screen Shot 2019-07-15 at 18 30 12

I didn't like the idea of constraining it to the Variation label. Open to any other suggestions.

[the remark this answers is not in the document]
ermau: Even if this was the correct change (not sure), you only applied it to one editor, but most have this same layout. I'll see if I can come up an alternative.

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantWindow.cs
```diff
@@ -0,0 +1,97 @@
...
+				NSLayoutConstraint.Create (createVariantView, NSLayoutAttribute.Height, NSLayoutRelation.Equal, ContentView, NSLayoutAttribute.Height, 1f, 0f),
+				NSLayoutConstraint.Create (createVariantView, NSLayoutAttribute.Width, NSLayoutRelation.Equal, ContentView, NSLayoutAttribute.Width, 1f, 0f),
+
+			});
+
+			var buttonOK = new FocusableCommandButton {
+				BezelStyle = NSBezelStyle.Rounded,
+				KeyEquivalent = "\r",
+				Command = ViewModel.CreateVariantCommand,
+				ControlSize = NSControlSize.Regular,
```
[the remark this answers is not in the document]
CartBlanche: It shouldn't be Regular. Removed.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
CartBlanche: Removed.

vancura:
@ermau Yes, there are rollovers. Everything is defined and clearly visible in the [Constraints Figma file](https://www.figma.com/file/tzh2CxiRxYqMNi3gNKpLg7/Property-Pad-%E2%80%93-Variations?node-id=0%3A1).

image

@CartBlanche:

> I don't have the bitmaps you speak of.

Same as above, they're in the file above. You're a UI engineer, you should be able to export from the document (referenced above and in many channels all around), in any possible resolution and format. I am not blocking you, Figma is straightforward, and you should be able to do these things without me.

Please make sure the appearance is pixel-perfect with my designs. Currently, they are far from how designed, and this needs to be fixed.

image

CartBlanche:
Mouse Overs...
Screen Shot 2019-08-14 at 12 24 34
Screen Shot 2019-08-14 at 12 25 01

vancura:
Looks perfect! Thanks!

CartBlanche:
@ermau In case we can't have that meeting the basic summary of the tab focus issue is...
1. I've made sure the EditorContain can become a `CanBecomeKeyView`
https://github.com/xamarin/Xamarin.PropertyEditing/pull/600/files#diff-46790649fe41a28c66ee28dbbf954c6dR88

2. When the `Delete` button is created I make it the `NextKeyView`
https://github.com/xamarin/Xamarin.PropertyEditing/pull/600/files#diff-46790649fe41a28c66ee28dbbf954c6dR157

3. When the `Add` button is created I make it the `NextKeyView`
https://github.com/xamarin/Xamarin.PropertyEditing/pull/600/files#diff-46790649fe41a28c66ee28dbbf954c6dR194

The FocusableCommandButton’s `CanBecomeKeyView` property fires
https://github.com/xamarin/Xamarin.PropertyEditing/pull/600/files#diff-621d5923d3c8cecf8297bde5de504b18R114

But it's `BecomeFirstResponder` method never fires...
https://github.com/xamarin/Xamarin.PropertyEditing/pull/600/files#diff-621d5923d3c8cecf8297bde5de504b18R128

I cannot see why, right now.

ermau:
I think we might want to add another property that the auto-focus on property button press thing uses to focus. Default it to the first key view and make it virtual. For variant properties, the remove variant button is the thing that gets focused after pressing the property button and that doesn't seem like the ideal behavior.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -63,6 +95,209 @@ public override void ViewWillMoveToSuperview (NSView newSuperview)
...
+			foreach (CALayer variationLayer in this.variationChildLayersList) {
+				if (variationLayer is CAShapeLayer lineShape) {
+					lineShape.StrokeColor = HostResources.GetNamedColor (NamedResources.ForegroundColor).CGColor;
+				}
+			}
+		}
+
+		private void UpdateVariations ()
+		{
+			ClearSubViews ();
```
ermau: You have checks to only create if there aren't buttons, but you're erasing the buttons here, recreating them regardless.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -63,6 +95,209 @@ public override void ViewWillMoveToSuperview (NSView newSuperview)
...
+
+			FrameChanged ();
+		}
+
+		protected void AppearanceChanged ()
+		{
+			// Repaint those lines
+			foreach (CALayer variationLayer in this.variationParentLayersList) {
+				if (variationLayer is CAShapeLayer lineShape) {
+					lineShape.StrokeColor = HostResources.GetNamedColor (NamedResources.ForegroundColor).CGColor;
```
ermau: You could grab this color once and reuse it.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -63,6 +95,209 @@ public override void ViewWillMoveToSuperview (NSView newSuperview)
...
+				}
+			}
+		}
+
+		private void UpdateVariations ()
+		{
+			ClearSubViews ();
+
+			if (this.viewModelAsPropertyViewModel != null && this.viewModelAsPropertyViewModel.Property.CanWrite) {
```
ermau:
```suggestion
			if (this.viewModelAsPropertyViewModel != null && this.viewModelAsPropertyViewModel.IsInputEnabled) {
```

Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
```diff
@@ -165,6 +165,7 @@ private void RequireComboBox()
 			AddSubview (this.comboBox);
 
 			AddConstraints (new[] {
+				NSLayoutConstraint.Create (this.comboBox, NSLayoutAttribute.Bottom, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Bottom, 1f, BottomOffset),
```
ermau: Shouldn't this be in place of the CenterY constraint?

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -157,5 +165,18 @@ internal new TViewModel ViewModel
...
 		}
+
+		public override bool IsDynamicallySized => true;
+
+		public override nint GetHeight (EditorViewModel vm)
+		{
+			nint baseHeight = base.GetHeight (vm);
+			var realVm = (PropertyViewModel)vm;
+			if (realVm != null && realVm.IsVariant) {
+				return baseHeight * 2;
```
ermau: Why is it multiplied? This is just going to lead to uneven sizes when trying to adjust the multiplier for different base sizes. Just add the fixed size of the variant header.

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantView.cs
```diff
@@ -0,0 +1,124 @@
...
+			Initialize (createVariantViewModel, windowWidthHeight);
+		}
+
+		private void Initialize (CreateVariantViewModel createVariantViewModel, CGSize windowWidthHeight)
+		{
+			Frame = new CGRect (CGPoint.Empty, windowWidthHeight);
+
+			int editorHeight = 18;
+
+			var FrameWidthThird = Frame.Width / 3;
```
ermau: Shouldn't this be the same sizing as the panel?

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantView.cs
```diff
@@ -0,0 +1,124 @@
...
+				}
+
+				controlTop += editorHeight + HorizontalControlSpacing;
+			}
+
+			Frame = new CGRect (CGPoint.Empty, new CGSize (Frame.Width, controlTop));
+		}
+	}
+
+	internal class VariationFacade
```
ermau: This could be private to the view.

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantWindow.cs
```diff
@@ -0,0 +1,95 @@
...
+			if (propertyViewModel == null)
+				throw new ArgumentNullException (nameof (propertyViewModel));
+
+			Delegate = new ModalWindowCloseDelegate ();
+
+			FloatingPanel = true;
+
+			Title = Properties.Resources.AddVariationTitle;
+
+			MaxSize = new CGSize (minWindowWidth * 2, minWindowHeight * 2); // TODO discuss what the Max/Min Size should be and if we should have one.
```
ermau: I don't think we really need one

Xamarin.PropertyEditing.Mac/Controls/RatioEditorControl.cs
```diff
@@ -26,10 +26,9 @@ public RatioEditorControl (IHostResourceProvider hostResources)
 			};
 			AddSubview (this.ratioEditor);
 
-			this.AddConstraints (new[] {
-				NSLayoutConstraint.Create (this.ratioEditor, NSLayoutAttribute.CenterY, NSLayoutRelation.Equal, this,  NSLayoutAttribute.CenterY, 1f, 0),
+			AddConstraints (new[] {
+				NSLayoutConstraint.Create (this.ratioEditor, NSLayoutAttribute.Bottom, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Bottom, 1f, 0),
```
ermau: No constant offset?

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -157,5 +165,18 @@ internal new TViewModel ViewModel
 			get { return (TViewModel)base.ViewModel; }
 			set { base.ViewModel = value; }
 		}
+
+		public override bool IsDynamicallySized => true;
+
+		public override nint GetHeight (EditorViewModel vm)
```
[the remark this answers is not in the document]
CartBlanche: Agreed.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -157,5 +165,18 @@ internal new TViewModel ViewModel
 			get { return (TViewModel)base.ViewModel; }
 			set { base.ViewModel = value; }
 		}
+
+		public override bool IsDynamicallySized => true;
```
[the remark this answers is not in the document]
CartBlanche: Agreed.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -13,23 +14,34 @@ public class PropertyButton
 
 		NSMenu popUpContextMenu;
 
+		private bool creatingVariantWindow;
```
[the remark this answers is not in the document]
CartBlanche: Done.

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableButton.cs
```diff
@@ -10,7 +11,23 @@ public abstract class UnfocusableButton : NSImageView
 		public event EventHandler OnMouseLeftDown;
 		public event EventHandler OnMouseRightDown;
 
-		NSTrackingArea trackingArea;
+		private NSTrackingArea trackingArea;
+
+		private ICommand command;
+
+		public ICommand Command
```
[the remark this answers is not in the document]
CartBlanche: I must have used this initially for the Add/Delete Variant and then switched to normal buttons and forgot to remove this code. Removed now.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -11,27 +16,52 @@ public EditorContainer (IHostResourceProvider hostResources, IEditorView editorV
...
+				UpdateVariations ();
+
+				FrameChanged ();
+			}
+		}
+
+		private void PropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
+		{
+			if (this.viewModelAsPropertyViewModel != null) {
+				if (e.PropertyName == nameof (this.viewModelAsPropertyViewModel.HasVariantChildren)) {
```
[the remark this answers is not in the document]
CartBlanche: OK.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
```diff
@@ -63,6 +95,209 @@ public override void ViewWillMoveToSuperview (NSView newSuperview)
...
+			}
+			endLineShape.Path = endLinePath.ToCGPath ();
+			endLineShape.FillColor = null;
+			endLineShape.Opacity = 1.0f;
+			endLineShape.StrokeColor = HostResources.GetNamedColor (NamedResources.ForegroundColor).CGColor;
+			return endLineShape;
+		}
+
+		// the deleteButton's AddObserver fires this method when the observed property changes
+		public override void ObserveValue (NSString keyPath, NSObject ofObject, NSDictionary change, IntPtr context)
```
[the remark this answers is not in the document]
CartBlanche: Yes, but only if we don't handle our value.

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantWindow.cs
CartBlanche: OK.

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantView.cs
CartBlanche: Sure.

Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
CartBlanche: Yes.

Xamarin.PropertyEditing.Mac/Controls/RatioEditorControl.cs
CartBlanche: Actually this highlights a bug, where the Base control's height was manually being set. This should be fixed now and BottomOffset added.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
CartBlanche: @ermau Could you clarify what you mean by "add the fixed size of the variant header"?

ermau: The variant header (as in the list of variations for this variant) is a fixed or calculate-able size, just add that to the base height plus margins.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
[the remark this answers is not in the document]
ermau: Ah, it is in `GetView`, but I see it ends up called in `GetRowHeight` first. It doesn't seem drastically slower so we'll just move forward with this for now, we can revisit it later as we'll need to revise the size registration system.

Xamarin.PropertyEditing.Mac/Controls/Variations/CreateVariantView.cs
CartBlanche: The idea was that the VariationCategory Name would be take up the left 3rd of the initial window, while the Category drop downs would take up the remaining 2/3rds and expand.
As per...
Screen Shot 2019-10-02 at 10 06 52
I did think of making it follow Golden ratio resizing, as per our property list, but thought it might be overkill of this popup window.

Xamarin.PropertyEditing.Mac/Controls/EditorContainer.cs
CartBlanche: They need to be cleared, so have removed the checks.

Xamarin.PropertyEditing.Mac/Controls/PropertyEditorControl.cs
```diff
@@ -89,7 +91,13 @@ public void UpdateKeyViews ()
 				} while (row > 0 && ctrl == null);
 
 				if (ctrl != null) {
-					ctrl.LastKeyView.NextKeyView = FirstKeyView;
+					VariationButton vb = Superview.Subviews.OfType<VariationButton> ().FirstOrDefault ();
```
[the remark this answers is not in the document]
CartBlanche: Agreed. I'd like to do that, but think it should be done under a separate issue, assuming this works and doesn't affect performance.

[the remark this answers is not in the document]
ermau: Making outside changes to better enable new scenarios don't have to be separate issues, that's just going to accumulate tech debt because they'll get lost in the priority, especially when it wouldn't have taken that long to have just done it that way in the first place.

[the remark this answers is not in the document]
CartBlanche: I'd totally focused on the 2nd. Just implemented the 1st.

CartBlanche:
Due to time, I've added a separate issue with regard to the autofocus property button - https://github.com/xamarin/Xamarin.PropertyEditing/issues/652

ermau:
Due to time? This doesn't have a desirable behavior, it is not a separate issue.

CartBlanche:
By Due to time, I talking about what Bill said. Also that desirable behaviour would be required if Variations was shipping, which at the moment, my understanding is that it is on hold. Hence why I set up the a separate issue. So that if it if we need to work it again, there is already an issue for it. Also it isn't that the current behaviour is unusable. It makes sense when tabbing. The other scenario is what needs enhancing.

ermau:
Variations are also intended for Form's use, they are outside of the scope of the discussion with Bill.

Xamarin.PropertyEditing.Mac/Controls/Custom/PropertyButton.cs
```diff
@@ -154,8 +164,8 @@ private void PopUpContextMenu ()
 
 		private void FocusClickedRow ()
 		{
-			if (Superview != null) {
-				MakeFocusableKeyViewFirstResponder (Superview.Subviews);
+			if (Superview is EditorContainer ec) {
```
CartBlanche: By only passing the `EditorView.NativeView.Subviews` the recursion only traverses down the `EditorContainer` tree, thus avoiding the `Variation` buttons.

PR #602, opened 2019-05-28
[loc] Add translations for our strings

alanmcgovern:


alanmcgovern:
@viniciusjarina @ermau can we keep proppy up to date with translations the same as we keep designer up to date? https://github.com/xamarin/designer#translations

PR #603, opened 2019-06-04
[Win] Multi-handler events

ermau:
Event editors can now indicate support for handling multiple handlers per event and provide suggested handler names.

image
image


Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
```diff
@@ -458,6 +458,76 @@
...
+		<Setter Property="ContentTemplate">
+			<Setter.Value>
+				<DataTemplate>
+					<Grid RenderTransformOrigin="0.5,0.5" Width="6" Height="6" VerticalAlignment="Center" HorizontalAlignment="Center">
+						<Grid.RenderTransform>
+							<RotateTransform Angle="45" />
+						</Grid.RenderTransform>
+
+						<Line X1="0" X2="6" Y1="3" Y2="3" Stroke="{Binding RelativeSource={RelativeSource AncestorType=Button},Path=Foreground}" StrokeThickness="1" />
+						<Line X1="3" X2="3" Y1="0" Y2="6" Stroke="{Binding RelativeSource={RelativeSource AncestorType=Button},Path=Foreground}" StrokeThickness="1" />
```
vancura: Could you please make the cross larger, so it looks like this?

image

image

Here it is `6.35pt` large. So something around that.

With `8pt` space from the label.

image

CartBlanche:
@ermau Noticed that we allow spaces within the event name...
Screen Shot 2019-07-10 at 10 34 14
Shouldn't we strip these out?

CartBlanche:
Also from an Accessibility perspective once an event is hooked up, I seem to be able to tab into that row, but there is no visual cue that I'm in there (no focus ring/rect) and I can't hit `Delete` to remove the event handler.

ermau:
> Shouldn't we strip spaces and invalid chars out?

No but we should provide an API for validation.

CartBlanche:
@ermau how are multiple event handlers added. Once I add 1 I have no dropdown to select a suggested one (as per recent screenshot) nor can I type another one in there.

ermau:
@CartBlanche go to `MockObjectEditor` and set `SupportsMultipleHandlers` to `true`.

CartBlanche:
@ermau Will you be looking at the Accessibility stuff in another PR?

CartBlanche:
Just noticed the `AddHandler` test is failing both locally and on CI.

vancura:
Can you paste final screenshots please, both light and dark themes?

ermau:
@vancura 

Xamarin PropertyEditing Windows Standalone_2019-07-25_14-27-52
Xamarin PropertyEditing Windows Standalone_2019-07-25_14-27-58

ermau:
@vancura 
image

Keep in mind the hitbox of the x button is a bit bigger than the actual x, currently the text and x share the same padding at 5pt, are you sure you want to put the hitbox at a different distance from the right edge than the text is?

vancura:
I don't know what's the hitbox, it's not visible and it's perfectly fine, and clickable. All I care about now is that the × symbol has the same distance from right as from top and bottom.

ermau:
@vancura It does become visible during rollover.

image

vancura:
Yup, correct, but as you can see the horizontal distance between the button and the right edge is pretty visible.

PR #619, opened 2019-08-28
[Win] Improve accessibility

ermau:


CartBlanche:
@ermau code looks fairly straight forward, but wondering about the best way to test the functionality?

ermau:
@CartBlanche Properties that previously only said what kind of editor you were in and the contents should now also list the property name. The accessibility inspector tool will also show the property as a group that contains the editor instead of a flat hierarchy.

PR #620, opened 2019-08-29
[Mac] Since the centre ellipsis works well, we no longer need label t…

CartBlanche:
…ooltips.


ermau:
It still provides an alternate way to get the full property name without just expanding the property panel. I see no reason to remove it.

CartBlanche:
@ermau You don't need to expand anything. If you mouse over a centre ellipsis label it automagically (no extra code required) pops up a tooltip, over the compressed text and expands the tooltip to the show the full text. 
This really is redundant now, as it shows tooltips for labels which are always *clearly* visible.

This is what the expanded tooltip on centre ellipsis text looks like...
Screen Shot 2019-08-29 at 17 03 43

Issue #616, opened 2019-08-27
Windows Auto-resizing editor


ermau:
![image](https://user-images.githubusercontent.com/156582/63870770-a4417800-c988-11e9-9d3d-9d9d9b11d99e.png)
![image](https://user-images.githubusercontent.com/156582/63870843-c804be00-c988-11e9-9b73-5c1a70ce4cd8.png)

vancura:
There are a couple of issues, especially in different spaces around the inner rectangle, lengths of each line and mainly about stroke width (it's centered `1pt` and hence antialiased, semi-transparent). I'd like to properly design this editor, do we have time?

ermau:
@vancura Only if you can get it to me this week and it doesn't represent a huge departure. I'll address the centering issue, I just wanted to get a photo off to you to hopefully get some colors back quickly.

vancura:
What do you think about this?

![proppy@2x](https://user-images.githubusercontent.com/4982/64353329-de9cbc00-cffd-11e9-9342-4233d59a6ce6.png)

Unfortunately we don't have the Autoresizing and Origin colors in the `ColorService` library, you'll need to create them in code or add new tokens.

### Mac + Windows light theme colors:
- grayscale dots and guidelines: #555555
- red: #F56D4F

### Mac + Windows dark theme colors:
- grayscale dots and guidelines: #555555
- red: #F56D4F (same as everywhere)

### Windows dark theme colors:
- grayscale dots and guidelines: #777777
- red: #F56D4F (same as everywhere)

Thanks!

ermau:
@vancura looks good to me, thanks!

CartBlanche:
@vancura That looks good! We also need colours for when the control is disabled? ATM I'm converting the "active" colours (the reds) to gray (different to your aforementioned grays) and leaving the rest as is. Looks ok, but something more inline with your design would be better, IMO.

vancura:
The standard way is to make the control 50% opacity when disabled. Can you test this? I think that should be fine.

PR #630, opened 2019-09-16
[Mac] Autoresizing implementation

CartBlanche:
Screen Shot 2019-10-10 at 20 40 54


CartBlanche:
@viniciusjarina & @chkn This is the port of the Autoresizing control into proppy. Could you please look over it and the video I posted on Slack in #proppy (https://xamarinhq.slack.com/archives/C5GBTFXL0/p1568909239055500). It Supports theming and keyboard navigation (Left and Right arrows to cycle through the flags). With theming we need to decide what the actual colour scheme will be between dark and light, but that should be easily changed once we finalise what that will actually be.

ermau:
@CartBlanche Theming is here https://github.com/xamarin/Xamarin.PropertyEditing/issues/616#issuecomment-528410763

ermau:
You'll want to rebase this and use the rectangle sizing methodology now present in the view model as it is more correct.

CartBlanche:
@ermau just to be clear, you are referring to `VM.GetPreviewElementRectangle`?

ermau:
@CartBlanche Correct.

CartBlanche:
I made it a vertical alignment as it really is better aesthetically than the horizontal one that doesn't support resizing. We can always revisit the alignment, should Forms be needed.

PR #633, opened 2019-09-24
Stateful properties

ermau:
This adds stateful properties for Windows. The design and precise implementation can still be improved upon, but this establishes the integration points and serves as a working implementation.

Category arrange:

Named arrange:


CartBlanche:
@ermau While using it I'm wondering if the sub-items in `Category` mode should be expanded by default?

CartBlanche:
It also needs a rebase.

PR #634, opened 2019-09-24
Fix UpdateResourcesAsync crash.

viniciusjarina:
* Changed do while for while, to avoid crash with WhenAny in case of the HashSet is empty.


ermau:
This isn't a correct fix and just hides a bug somewhere else. There should be no circumstances in which this is created without any targets. We should add an argument check for the size of the targets array in `ctor` and fix whatever is causing this to be created with an empty array.

PR #641, opened 2019-10-01
[mac] Ignore change events if no resources

ermau:
Seems like we might be getting chang events even when not visible so let's ignore them
if we don't actually have a resource selector. The search box gets added regardless to
simplify putting it at the end.


garuma:
@monojenkins backport to d16-4

PR #645, opened 2019-10-03
Adds a mechanism in PredefinedValues to allow handle separator string values

### LINKED ISSUE #644: [Feature] Include a mechanism to include separators in PredefinedValues
In VS4Mac we have some PredefinedValues that include dividers to make more clear the selection of items for the user.

image

I did an approach here https://github.com/xamarin/Xamarin.PropertyEditing/commit/5d8f2e2e625bac39fa47d1ed1533ceac51af1cda  @ermau what do you think?

image

netonjm:
In VS4Mac we have some PredefinedValues that include dividers to make more clear the selection of items for the user.

image

This is how it feels now.

image

Fixes #644

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
@@ -39,14 +39,22 @@ namespace Xamarin.PropertyEditing.Mac
 
 				this.popupButtonList.RemoveAllItems ();
 				foreach (var item in ViewModel.PossibleValues) {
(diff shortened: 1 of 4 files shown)


netonjm:
@ermau what do you think? (btw I didn't include windows changes implementing the interface)

CartBlanche:
@netonjm FYI - You can set PRs as **Draft** now and that way they can't be merged until you change them to normal PRs. That way you don't need the "do no merge" tag.

ermau:
Probably going to have to do something like this for iOS, so we might need something a little more robust.

PR #647, opened 2019-10-03
[Core] Don't assume default is default(TValue)

ermau:
This removes a seeming requirement that the default value of a
predefiend values constrained property be the default of the type.

A object editor that properly supports default will return a default
value that is in the list of predefined values. Resetting the value does
not attempt to use the default value and instead simply passes Unset, so
there's no need for the default value to need to be known.


netonjm:
LGTM!
image

PR #649, opened 2019-10-04
Windows Autoresizing Editor

ermau:


CartBlanche:
I just did a quick test on Windows and noticed that if you resize smaller than the preview window, you get a weird overlap. As per screenshot...
Screen Shot 2019-10-07 at 12 38 18

If we are supporting preview resize, should both the editor and preview resize accordingly?
Does it have to resize in v1.0?

ermau:
The design is going to have to be changed in some way to support scaling, it doesn't really fit the current minimum size. Let's get it in so it doesn't rot and we can address it later if it actually gets used.

CartBlanche:
I actually think in our proppy, it would be better if the Editor and Previewer were aligned vertically. Makes for a taller row, but less need for scaling.

PR #658, opened 2019-10-11
[Mac] Fix crash on Catalina by using NSOpenPanel.OpenPanel static met…

### LINKED ISSUE #657: Potential crash with NSOpenPanel in BasePathEditorControl on Catalina
With Catalina NSSavePanel panels run out-of-proc so that instantiation is not supported anymore (in fact it seems that it was never officially supported) and will crash, hence we have to use the static shared panels instead (`NSSavePanel.SavePanel` or `NSOpenPanel.OpenPanel`).

For more details see: https://github.com/xamarin/xamarin-macios/issues/6474

Affected code: https://github.com/xamarin/Xamarin.PropertyEditing/blob/38105d0dcc46082d5d249de92ab188d829ea4dc1/Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs#L43-L49

Bug filed for VS for Mac: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/989519

CartBlanche:
…hod.

Fixes #657 

Step to repro

1. Clone repo
2. Should be in master
3. Build and run the stand alone project.
4. Go to the FilePath editor and click the ellipsis. 
5. You should see the reported crash.
6. Switch to dominique-Fix657 branch
7. Repeat steps 1-5 and you shouldn't have a crash.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
@@ -40,13 +40,7 @@ namespace Xamarin.PropertyEditing.Mac
 
 			this.currentTextField.AddButton (this.revealPathButton);
 
-			this.panel = new NSOpenPanel {
-				AllowsMultipleSelection = false,
-				CanCreateDirectories = true,
-				ShowsHiddenFiles = false,
-				ShowsResizeIndicator = true,
-				TreatsFilePackagesAsDirectories = true,
-			};
+			this.panel = NSOpenPanel.OpenPanel;
 
 			// update the value on keypress
 			this.revealPathButton.Activated += OnRevealPathButtonActivated;
@@ -92,7 +86,13 @@ namespace Xamarin.PropertyEditing.Mac
 		private void BrowsePathButton_Activated (object sender, EventArgs e)
 		{
 			Window.MakeFirstResponder (this.currentTextField);
-			this.panel.BeginSheet (this.Window, HandleAction);
+
+			this.panel.AllowsMultipleSelection = false;
+			this.panel.CanCreateDirectories = true;
+			this.panel.ShowsHiddenFiles = false;
+			this.panel.ShowsResizeIndicator = true;
+			this.panel.TreatsFilePackagesAsDirectories = true;
+			this.panel.BeginSheet (Window, HandleAction);
 		}
 
 		protected abstract void OnRevealPathButtonActivated (object sender, EventArgs e);


CartBlanche:
@sevoku I don't think that would happen as we are showing it modally, so the user shouldn't be able to pull open another panel, or am I missing something?

sevoku:
nono, what I mean is this: https://github.com/xamarin/Xamarin.PropertyEditing/blob/962ddaab047f2570161ff3911334e199ec999f19/Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs#L43-L48 is in the constructor. So the control gets created, maybe several times and sits in the grid until the user clicks the button, right?

What if something else (like in VS Mac context, simply the open/save file command) will use the shared panel in-between (so after that has been set in the constructor, but before the button gets activated) and modify the same properties to its needs. Now if the user clicks the button and you show the panel, it might have completely different filters, an accessory view and whatever else.

CartBlanche:
OK I'm just building a local copy of Vs4m, so can test that scenario soon.

CartBlanche:
@sevoku I don't think this static method is returning a shared OpenPanel. It is simply giving us a standard way of creating an NSOpenPanel, while Apple, hide the implementation specifics. So I'd be very surprised if 2 panels within the same app, are sharing the same instance. IT is definitely worth testing to be 100% sure, but I really would be surprised if that happens.

CartBlanche:
@sevoku Assuming this fixes the crash, the I think we need to raise another issue to fix the UX aspect you mention above. Right now the crash should be the priority, IMO.

garuma:
Buddy tested the changes on my side on Catalina.

I wasn't able to get the standalone to crash on master, discussing this on #proppy channel it seems like the crash could only be present under certain circumstances (e.g. only in bundles that are hardened/notarized). I tested the changes of this PR and since it doesn't change anything in behavior I would recommend we just merge it

Xamarin.PropertyEditing.Mac/Controls/BasePathEditorControl.cs
```diff
@@ -92,7 +86,13 @@ private void CurrentTextField_Changed (object sender, EventArgs e)
 		private void BrowsePathButton_Activated (object sender, EventArgs e)
 		{
 			Window.MakeFirstResponder (this.currentTextField);
-			this.panel.BeginSheet (this.Window, HandleAction);
+
+			this.panel.AllowsMultipleSelection = false;
```
CartBlanche: This was the properties required for Browse to work will always be set correctly, before we open the panel.

CartBlanche:
I also just checked and each proppy control receives a new instance of the `NSOpenPanel`. The objects aren't shared between controls. So presumably if VS4M is creating it's own via `NSOpenPanel.OpenPanel`, it will also have a separate instance too.

PR #674, opened 2019-10-24
[Mac] Initial Events implementation.

CartBlanche:


CartBlanche:
@mcumming did someone else end up implementing this or has the Property Editor been replaced?

PR #675, opened 2019-10-25
Adds manual focus chain to views instead expect automatic behaviour from window

netonjm:
In VS4Mac the autocalculation of the keyviewloop chain is not well done by the gtkquarz window, and because this we need to ensure NextKeyView are correctly set in the views.

9DUbofnfVY

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableScrollView.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableScrollView.cs
@@ -1,0 +1,9 @@
+﻿using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	class UnfocusableScrollView : NSScrollView
+	{
+		public override bool AcceptsFirstResponder () => false;
+	}
+}
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableStackView.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableStackView.cs
@@ -1,0 +1,9 @@
+﻿using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	class UnfocusableStackView : NSStackView
+	{
+		public override bool AcceptsFirstResponder () => false;
+	}
+}
--- a/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
@@ -105,6 +105,7 @@ namespace Xamarin.PropertyEditing.Mac
 				if (this.viewModel != null) {
 					this.viewModel.PropertyChanged += OnVmPropertyChanged;
 
+					TabButton lastArrangeMode = null;
 					for (int i = 0; i < this.viewModel.ArrangeModes.Count; i++) {
 						var item = this.viewModel.ArrangeModes[i];
 						string imageName = GetIconName (item.ArrangeMode);
@@ -121,6 +122,14 @@ namespace Xamarin.PropertyEditing.Mac
 						arrangeMode.AccessibilityTitle = string.Format (Properties.Resources.ArrangeByButtonName, item.ArrangeMode.ToString());
 
 						this.tabStack.AddView (arrangeMode, NSStackViewGravity.Top);
+
+						if (i == this.viewModel.ArrangeModes.Count - 1) {
+							arrangeMode.NextKeyView = propertyFilter;
+						}
+						if (lastArrangeMode != null) {
+							lastArrangeMode.NextKeyView = arrangeMode;
+						}
+						lastArrangeMode = arrangeMode;
 					}
 				}
 			}
@@ -182,6 +191,7 @@ namespace Xamarin.PropertyEditing.Mac
 				TranslatesAutoresizingMaskIntoConstraints = false
 			};
 			AddSubview (this.propertyList);
+			propertyFilter.NextKeyView = propertyList;
 
 			this.AddConstraints (new[] {
 				NSLayoutConstraint.Create (this.header, NSLayoutAttribute.Top, NSLayoutRelation.Equal, this, NSLayoutAttribute.Top, 1, 0),
--- a/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
@@ -13,6 +13,8 @@ namespace Xamarin.PropertyEditing.Mac
 {
 	public partial class PropertyEditorPanel : NSView
 	{
+		public override bool AcceptsFirstResponder () => false;
+
 		public PropertyEditorPanel ()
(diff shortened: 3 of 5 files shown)


Xamarin.PropertyEditing.Mac/PropertyList.cs
```diff
@@ -119,6 +122,16 @@ public override bool ValidateProposedFirstResponder (NSResponder responder, NSEv
 				return true;
 			}
 
+			private const string disclosureKey = "NSOutlineViewDisclosureButtonKey";
+			public override NSView MakeView (string identifier, NSObject owner)
+			{
+				NSView view = base.MakeView (identifier, owner);
+				if (identifier == disclosureKey && view is NSButton btn) {
+					btn.RefusesFirstResponder = true;
```
CartBlanche: Why don't we want the disclosure button to not accept focus? How will those with accessibility issue expand and contract nodes?

netonjm: because they are not included in your NextKeyLoop logic, then it brokes the current focus behaviour. we can, remove this code, and add it again when this will fixed. 

Any way as we were talking the other day, in VS4Mac section headers are not focusable, like for example the Toolbox. People with disabilities can collapse this regions using keyboard using VoiceOver

netonjm:
guys I would like to bump proppy submodule with this fixes for our EOD tomorrow

Xamarin.PropertyEditing.Mac/PropertyList.cs
sevoku: This makes sense, let's keep them disabled until we can fix the order properly. @CartBlanche the focus order is broken anyways, so it's not very useful to allow expanders to receive focus and since expanders are not always keyboard accessible on Mac in general, the official way to navigate those is to use VO, we have also an expmption on this for MAS-C.

CartBlanche:
@ermau thoughts on the disclosure tabbing, mentioned above?

Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
```diff
@@ -180,6 +191,7 @@ private void Initialize ()
 				TranslatesAutoresizingMaskIntoConstraints = false
 			};
 			AddSubview (this.propertyList);
+			propertyFilter.NextKeyView = propertyList;
```
CartBlanche: for consistency, we should prefix both `propertyFilter` and `propertyList` with `this.`

Xamarin.PropertyEditing.Mac/PropertyList.cs
ermau: While VSM has all categories expanded by default, this isn't the case for the other platforms. If the expander isn't focusable, I don't see how you would be able to expand or collapse it via keyboard.

netonjm: The problem @ermau is right now the groups disclosures are focusable but you didn't include it into your NextKeyView mechanism. This means the window calculates what's the next/previous view for this element. 
And the problem comes pressing tab from SearchTextField or any view without this, in my last tests in VisualStudio for Mac the autocalculation from the window was resolving to focus the  groups disclosures and as we have said before they are not set which ends in a broken statement and the user cannot change the focus over this element.

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableStackView.cs
```diff
@@ -0,0 +1,9 @@
+﻿using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	class UnfocusableStackView : NSStackView
```
Therzok:
```suggestion
	sealed class UnfocusableStackView : NSStackView
```

Xamarin.PropertyEditing.Mac/Controls/Custom/UnfocusableScrollView.cs
```diff
@@ -0,0 +1,9 @@
+﻿using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	class UnfocusableScrollView : NSScrollView
```
Therzok:
```suggestion
	sealed class UnfocusableScrollView : NSScrollView
```

Xamarin.PropertyEditing.Mac/PropertyList.cs
```diff
@@ -13,6 +13,9 @@ internal class PropertyList
 	{
 		internal const string PropertyEditorColId = "PropertyEditors";
 
+		public override bool AcceptsFirstResponder () => true;
+		public override bool BecomeFirstResponder () => Window.MakeFirstResponder (propertyTable);
```
Therzok:
```suggestion
		public override bool BecomeFirstResponder () => Window?.MakeFirstResponder (propertyTable);
```

CartBlanche: This will throw a compile error due to the possibility of a Window being null, which isn't compatible with the return type of bool. We'd need a bool?

Xamarin.PropertyEditing.Mac/PropertyList.cs
sevoku: @ermau sorry for the late response. That's the point, not everything is accessible in macOS in general. To be clear: unlike on Windows, keyboard navigation is not an accessibility feature on mac, it's made for convenience first of all, with the main focus on alerts/dialogs where the user is supposed to be able to to quickly tab though some workflow. i.e. if you try tabbing inside Finder or a more sophisticated UI like the Xcode main window, you'll find many dead ends, unreachable areas and keyboard traps. The official accessible way to navigate UIs with alternative input methods (non-pointer) is using VO navigation, which has a bunch of different settings incl. keyboard and it's even possible to disable the output and use it for navigation only. Also note that the Tab-Key setting is found in "System Prefs > Keyboard > Shortcuts" and not "System Prefs > Accessibility". For the same reason we have the exemption on keyboard navigation in VSMac.

ermau: @sevoku You can expand and collapse folders in Finder via keyboard navigation, I don't see this as any different. We will pursue being able to do this.

netonjm: I think the point here is: or we fix the current implementation including it in the current NextKeyView loop (current implementation is not doing it), or if not we should make it not focusable to don’t create breaking keyloop statements

sevoku: @ermau you didn't get my point: although it'd be great to support that, we don't need that to get MAS-C, that's it. From a11y terminology perspective it's a usability and not a cimpliance relevant issue. If we can fix it now, perfectly fine. If we need to work it out and put compliance by end of year at risk, pleas take the workaround or whatever.

Xamarin.PropertyEditing.Mac/PropertyList.cs
netonjm: Window?.MakeFirstResponder (propertyTable) ?? false

PR #680, opened 2019-11-07
[mac] Use more reliable row removed notice

ermau:
This method seems far more reliable for noticing when rows are removed so we can use it instead of changing to null superview (which sometimes it doesn't do). This fixes 966531.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
@@ -184,6 +184,13 @@ namespace Xamarin.PropertyEditing.Mac
 			}
 		}
 
+		public override void DidRemoveRowView (NSOutlineView outlineView, NSTableRowView rowView, nint row)
+		{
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -184,6 +184,13 @@ public override void ItemDidCollapse (NSNotification notification)
 			}
 		}
 
+		public override void DidRemoveRowView (NSOutlineView outlineView, NSTableRowView rowView, nint row)
+		{
+			if (rowView.Subviews[0] is EditorContainer ec) {
+				ec.ViewModel = null;
+			}
+		}
```
CartBlanche: As we are overriding, don't we want to call `base.DidRemoveRowView()` too?

ermau: There is no base for native types like this, if you call it it will throw.

Issue #684, opened 2019-11-13
[Mac] After Popping up a PopOver from Object, and the PopOver is cancel, focus should stay on the button 


CartBlanche:
https://devdiv.visualstudio.com/DevDiv/_queries/query/f37416fe-e5cb-4340-95f1-bb8d374b4d0d/

PR #685, opened 2019-11-13
[Mac] Make the button that popped up the PopOver have focus, once the…

CartBlanche:
… popover closes.

Fixes https://github.com/xamarin/Xamarin.PropertyEditing/issues/684

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
@@ -1,10 +1,7 @@
 using System;
-using System.Collections.Generic;
-using System.Linq;
 
 using AppKit;
 using CoreGraphics;
-using Foundation;
 
 using Xamarin.PropertyEditing.ViewModels;
 
@@ -16,7 +13,12 @@ namespace Xamarin.PropertyEditing.Mac
 		public CollectionEditorWindow (IHostResourceProvider hostResources, CollectionPropertyViewModel viewModel)
 			: base (new CGRect (0, 0, 500, 400), NSWindowStyle.Titled | NSWindowStyle.Closable | NSWindowStyle.Resizable, NSBackingStore.Buffered, true)
 		{
-			Delegate = new ModalDialogDelegate ();
+			if (hostResources == null)
+				throw new ArgumentNullException (nameof (hostResources));
+			if (viewModel == null)
+				throw new ArgumentNullException (nameof (viewModel));
+
+			Delegate = new ModalWindowCloseDelegate ();
 			Title = String.Format (Properties.Resources.CollectionEditorTitle, viewModel.Property.Name);
 
 			this.collectionEditor = new CollectionEditorControl (hostResources) {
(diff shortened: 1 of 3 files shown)


CartBlanche:
@netonjm I think this PR may fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1002094 .

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
```diff
@@ -26,19 +26,28 @@ public CollectionEditorWindow (IHostResourceProvider hostResources, CollectionPr
 
 			ContentView.AddSubview (this.collectionEditor);
 
-			this.ok = NSButton.CreateButton (Properties.Resources.OK, OnOked);
-			this.ok.AccessibilityEnabled = true;
-			this.ok.AccessibilityTitle = Properties.Resources.AccessibilityCollectionOKButton;
-			this.ok.Highlighted = true;
-			this.ok.TranslatesAutoresizingMaskIntoConstraints = false;
-			
-			//this.ok.KeyEquivalent = "\r"; // FIXME: The type selector popup doesn't eat this key, so it ends up closing both.
```
ermau: If you're going to uncomment/reintroduce a bit of code that has an explanation as to why it's commented out, be sure to re-test the stated issue. In this case, this broken behavior (pressing enter while the type selector is open closes the entire dialog) is now present again.

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
```diff
@@ -16,8 +16,8 @@ internal class CollectionEditorWindow
 		public CollectionEditorWindow (IHostResourceProvider hostResources, CollectionPropertyViewModel viewModel)
 			: base (new CGRect (0, 0, 500, 400), NSWindowStyle.Titled | NSWindowStyle.Closable | NSWindowStyle.Resizable, NSBackingStore.Buffered, true)
 		{
-			Delegate = new ModalDialogDelegate ();
-			Title = String.Format (Properties.Resources.CollectionEditorTitle, viewModel.Property.Name);
+			Delegate = new ModalWindowCloseDelegate (); //new ModalDialogDelegate ();
```
ermau: Replacing this usage with a common class that does the same thing is great, but let's remove the old usage and the old class if we're not going to use them.

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
```diff
@@ -16,8 +16,8 @@ internal class CollectionEditorWindow
...
+			Title = string.Format (Properties.Resources.CollectionEditorTitle, viewModel.Property.Name);
```
ermau: Unwanted irrelevant change.

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
```diff
@@ -26,19 +26,28 @@ public CollectionEditorWindow (IHostResourceProvider hostResources, CollectionPr
...
 			ContentView.AddSubview (this.collectionEditor);
 
-			this.ok = NSButton.CreateButton (Properties.Resources.OK, OnOked);
-			this.ok.AccessibilityEnabled = true;
-			this.ok.AccessibilityTitle = Properties.Resources.AccessibilityCollectionOKButton;
-			this.ok.Highlighted = true;
-			this.ok.TranslatesAutoresizingMaskIntoConstraints = false;
-			
-			//this.ok.KeyEquivalent = "\r"; // FIXME: The type selector popup doesn't eat this key, so it ends up closing both.
+			this.ok = new FocusableButton {
```
ermau: These buttons are already focusable, what's the need to change it?

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
CartBlanche: Meant to do that, forgot.

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
CartBlanche: They aren't though. They don't allow the user to tab to them in `master`.

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
CartBlanche: Fixed.

Xamarin.PropertyEditing.Mac/Controls/ControlExtensions.cs
```diff
@@ -37,6 +37,9 @@ public static Task<ITypeInfo> RequestAt (this TypeRequestedEventArgs self, IHost
 			popover.SetAppearance (hostResources.GetVibrantAppearance (source.EffectiveAppearance));
 
 			tcs.Task.ContinueWith (t => {
+				// Focus back to the button that popped us up
+				source.Superview.Window.MakeFirstResponder (source);
```
Therzok: Is there any cancellation support? This can break, as nothing guarantees the view or window is still displayed/valid by the time the continuation is called.

CartBlanche: @Therzok Can you clarify what you mean by "cancelation support"?

Therzok: In example, tcs.Task.ContinueWith means that the continuation callback will run after task.Task is finished.

In case we have something like:
```csharp
source.Dispose();
await tcs.Task; // this will run the continuation callback
```

this will crash on executing source.Superview in the continuation above, simply because we've disposed the view in the meantime. A CancellationTokenSource should be bound to the view and cancelled when the view is disposed.

ermau: There really shouldn't ever be a scenario in which that happens. Worst case I'd say let's just safeguard the continuation, we don't need to set up a whole cancellation setup for this.

CartBlanche: @ermau are you talking about just ensuring `source` isn't null at this point or some other belt and braces?

ermau: The superview and the window. `source` can't be null, it's a closed over local that's never set to null.

CartBlanche: Done.

Xamarin.PropertyEditing.Mac/Controls/CollectionEditorWindow.cs
netonjm: Why this ? https://github.com/xamarin/Xamarin.PropertyEditing/blob/f2f3549c656b566a32ee52538298f06ce43190d3/Xamarin.PropertyEditing.Mac/Controls/Custom/FocusableButton.cs#L8

A disabled view cannot be KeyView by default, no?

Xamarin.PropertyEditing.Mac/Controls/ControlExtensions.cs
```diff
@@ -37,6 +37,9 @@ public static Task<ITypeInfo> RequestAt (this TypeRequestedEventArgs self, IHost
 			popover.SetAppearance (hostResources.GetVibrantAppearance (source.EffectiveAppearance));
 
 			tcs.Task.ContinueWith (t => {
+				// Focus back to the button that popped us up
+				source.Superview?.Window?.MakeFirstResponder (source);
```
netonjm: 👍

PR #697, opened 2020-01-22
[VS] Show focus via focus rect, not color, for choice buttons

BretJohnson:
This fixes AB#1003978
Previously the choice buttons (used for picking brush type) showed focus via a different background color. But that color difference wasn't sufficiently obvious, per a11y testing, especially in the blue theme. Our UI designer suggested that we should show a focus rect to fix this.

Old UI (blue theme; note that the 3rd button has the focus):
image

Updated UI (blue theme):
image

Updated UI (dark theme):
image

### DIFF
--- a/Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
+++ b/Xamarin.PropertyEditing.Windows/Themes/Resources.xaml
@@ -202,7 +202,7 @@
 		<Setter Property="BorderThickness" Value="1" />
 		<Setter Property="Padding" Value="4,2,4,2" />
 		<Setter Property="Margin" Value="1,0,0,0" />
-		<Setter Property="FocusVisualStyle" Value="{x:Null}" />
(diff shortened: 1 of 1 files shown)


BretJohnson:
Actually, XET uses a focus rect for some tab UI (e.g. for text style attributes) but for brush picker they change the background color. I'm going to circle back with David Culbertson, the UI designer, on this one & see what we want to do.

BretJohnson:
I discussed this more with David, in UX. And he decided that the best fix here is to show the focus rect in addition to the background color. Per David, "That way we preserve the visual appeal that was there, and simply augment it with a standard, extra visual that's familiar to keyboard users."

And the UWP/WPF designer should do the same, adding the focus rect. Here's the bug to track that: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1056526

Note that this issue is really worst in the blue theme, where contrast is poor. XET has the same issue with blue theme, so making some fix there too makes sense.

PR #704, opened 2020-01-31
[Mac] Allow Ctrl+Shift+i to display the Property Button's menu.

CartBlanche:


CartBlanche:
Not particularly happy about the hoops to jump through to trigger the pop-up, but this appears to work. If you know of a more elegant solution, shout out!

chkn:
Doesn't look too bad to me

CartBlanche:
@vancura Do you have a suggestion other than Ctrl+Space?

sevoku:
There is no shortcut for this on a Mac and we should not add it just for proppy and nowhere else. Also CTRL+SPACE is a Windows specific shortcut.

With VO there is VO+SHIFT+M to open a context menu of the current element. For this to work you might need to implement [accessibilityPerformShowMenu](https://developer.apple.com/documentation/appkit/nsaccessibilityprotocol/1532774-accessibilityperformshowmenu)

vancura:
@CartBlanche We have an ongoing conversation in our team room, I'll paste the results here when we have a solution.

vancura:
@CartBlanche We don't have a solution :) Please don't use that shortcut.

PR #705, opened 2020-02-14
[Mac] Make Material View Tabbable.

CartBlanche:


Xamarin.PropertyEditing.Mac/Controls/Custom/MaterialView.cs
```diff
@@ -139,55 +147,61 @@ MaterialColorLayer CreateLayer (CommonColor color)
...
+				};
+
+				accentColourButton.Activated += MaterialColourButton_Activated;
+
+				AddSubview (accentColourButton);
+
 				x += width;
 			}
+
+			Window?.RecalculateKeyViewLoop ();
```
chkn: `Window` is null here the first time this is called. See the commit I pushed to call it earlier

Xamarin.PropertyEditing.Mac/Controls/Custom/MaterialView.cs
```diff
@@ -32,26 +32,17 @@ private void Initialize ()
...
 		}
 
-		public override void Layout ()
+		private void CreateColourPallette ()
 		{
-			if (Layer?.Sublayers != null) {
-				foreach (var l in Layer.Sublayers) {
-					l.RemoveFromSuperLayer ();
-					l.Dispose ();
+			if (Subviews.Length > 0) {
```
Therzok: Capture subviews in a local, otherwise this marshals the `Subviews` array twice.

Xamarin.PropertyEditing.Mac/Controls/Custom/MaterialView.cs
```diff
@@ -139,55 +147,61 @@ MaterialColorLayer CreateLayer (CommonColor color)
...
-						case MaterialColorType.Palette:
-							var match = ViewModel.Palettes.First (palette => palette.MainColor == editor.BackgroundColor);
-							ViewModel.ColorName = match.Name;
-							break;
-						}
-						NeedsLayout = true;
-					}
-				}
+			switch (e.PropertyName) {
+				case nameof (ViewModel.ColorName):
```
Therzok: Really nice, didn't think this was possible.

Xamarin.PropertyEditing.Mac/Controls/Custom/MaterialView.cs
CartBlanche: Done.

Xamarin.PropertyEditing.Mac/Controls/Custom/MaterialView.cs
CartBlanche: Thanks!

CartBlanche:
@chkn As mentioned, I also see that tabbing issue for the brown onwards, but right now I’m at a loss why `RecalculateKeyViewLoop ();` works for all the other colours but not those ones. Any suggestions?

BretJohnson:
I saw this before too. I believe the difference has to do with whether there is one row of lightness choices at the bottom or two, but beyond that not sure of the root cause.

CartBlanche:
Since all the layers are now inside buttons, who request layers, I don't think we need this anymore...

https://github.com/xamarin/Xamarin.PropertyEditing/blob/aaceed2af9efb489a0c724df0f86584b0e8015a6/Xamarin.PropertyEditing.Mac/Controls/Custom/MaterialView.cs#L23-L31

PR #706, opened 2020-02-17
[Mac] Tabbing to OutlineView in Category Mode

CartBlanche:


BretJohnson:
I tested this some. For the most part it seemed to work well.
I just spotted two minor issues.

First, when tabbing from the search box to the first category, the twistee (for Brush below) isn't highlighted. The Brush category header does seem to have the focus, since when I hit space it expands/collapses it, but the purple highlight doesn't show to indicate that..

image


Second, when I hit tab below the selected category doesn't change. That is, I can't select the `Windows Only` nor blank category at the end for some reason; I can only select the first three. This may or may not be a bug.

image

CartBlanche:
@BretJohnson I'm definitely getting highlights on the 1st 3. Seems toggling the last 2 manually, inserts it into the KeyView loop. Investigating.

CartBlanche:
Still not getting very far with why it magically works for 3 controls, and the other 2 it doesn't work until you expand one of them.

CartBlanche:
@BretJohnson, @chkn Here's a video of the code running within VS4Mac - https://youtu.be/SatQFXLyrIY

Standalone seems to have a tabbing, issue, which isn't visible in our shipped products.

For this to work in VS4Mac some VS4M code, that overrides tabbing, needs to be removed. 
I'll have a separate PR for that ready, as soon as this is merged into master.

PR #708, opened 2020-02-18
Makes not constrained PredefinedValues allow filtering + set value on enter

netonjm:
On VS4Mac we have some enumeration fields that the user can add custom values.


CartBlanche:
@netonjm This PR seems to have slipped through the net, or am I missing something @ermau? If it is supposed to go in, could you please rebase as there are conflicts.

PR #717, opened 2020-04-28
Fix NRE when SelectionChanged is firing on a null object.

CartBlanche:
Fixes https://github.com/xamarin/Xamarin.PropertyEditing/issues/718

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
@@ -48,6 +48,7 @@ namespace Xamarin.PropertyEditing.Mac
 				foreach (var item in ViewModel.PossibleValues) {
 					this.comboBox.Add (new NSString (item));
 				}
+				this.comboBox.SelectionChanged += ComboBox_SelectionChanged;
 			}
 
 			base.OnViewModelChanged (oldModel);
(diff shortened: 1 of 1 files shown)


netonjm:
The fix seems to be ok for me, because it's a NRE only to double check we don't crash we could also LOG in case of this.comboBox.SelectedValue is be null, which is a not expected situation

CartBlanche:
@netonjm `SelectedValue` can be `null`? When you 1st create the control and populate it. you have a list but SelectedIndex will be -1 and therefore value will also be null?

From the Apple docs: `If no item is selected, the value in this property is nil. Nothing is selected in a newly initialized combo box.`

CartBlanche:
But it is possible SelectedValue will be null if the control is reused, which seems to be the scenario we are hitting. Check put in for that.

Xamarin.PropertyEditing.Mac/Controls/PredefinedValuesEditor.cs
```diff
@@ -184,7 +184,9 @@ private void PopupButton_Activated (object sender, EventArgs e)
 
 		private void ComboBox_SelectionChanged (object sender, EventArgs e)
 		{
-			if (ViewModel != null && this.comboBox != null) {
+			if (ViewModel != null
+				&& this.comboBox != null
+				&& this.comboBox.SelectedValue != null) {
```
netonjm, line 189: && this.comboBox?.SelectedValue

CartBlanche: @netonjm It won't read the 3rd check if `this.combobox` is null. It will bail straight away because 1 of the 1st 2 checks has failed. The compiler usually checks the conditionals in sequence.

PR #723, opened 2020-06-08
[mac] fixes date control inconsistent

paymicro:
Context [AB#1137572](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1137572)
image

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
@@ -16,7 +16,8 @@ namespace Xamarin.PropertyEditing.Mac
 				DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond | NSDatePickerElementFlags.YearMonthDateDay,
 				DatePickerStyle = NSDatePickerStyle.TextFieldAndStepper,
 				Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Small)),
-				TranslatesAutoresizingMaskIntoConstraints = false
+				TranslatesAutoresizingMaskIntoConstraints = false,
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
```diff
@@ -16,7 +16,8 @@ public DateTimeEditorControl (IHostResourceProvider hostResources)
 				DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond | NSDatePickerElementFlags.YearMonthDateDay,
 				DatePickerStyle = NSDatePickerStyle.TextFieldAndStepper,
 				Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Small)),
-				TranslatesAutoresizingMaskIntoConstraints = false
+				TranslatesAutoresizingMaskIntoConstraints = false,
+				TimeZone = Foundation.NSTimeZone.FromAbbreviation ("GMT")
```
CartBlanche, line 20: Sorry about the late reply. Just wondering why we want to hard code this to "GMT"?  
Could you also give an example of the inconsistency you are seeing, so I have something to compare to?

netonjm: This should be UTC not GMT.


image

netonjm: This should be UTC, not GMT. Add the abbreviation into a constant.

9qYGQXRAjZ

netonjm: As the recent conversation this doesn't fix the real problem in Xam.Forms and it needs to be addressed in different way, adding some customisation to allow, in addition to full date time, only time or date.

PR #724, opened 2020-06-16
Ensure all the screen elements in this Window have their …

CartBlanche:
`AutomationProperties.Name` property set, for accessibility.

Fixes :  https://dev.azure.com/devdiv/DevDiv/_sprints/taskboard/Xamarin%20UI%20Tools/DevDiv/OneVS/CY20-H1/Q2/Sprint%20172?workitem=1004049

This has highlighted the need to run the Accessibility Insights for Windows tool over more of proppy, while it is running in VS. We will also need to do a pass to remove control type or similar.


Xamarin.PropertyEditing/Properties/Resources.resx
```diff
@@ -1,1047 +1,1057 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
decriptor, line 1: Did you mean to reformat this file?

CartBlanche: No :(

CartBlanche: @decriptor I used the Windows editor, which did that, I'll push a nicer fix :/

decriptor: I suppose though if that's how we want to do it for this repo a separate PR to "fix" the formatting might be good

CartBlanche: These are the least changes I could get right now. Even copying from master and adding the others on, messes up :/

CartBlanche:
@decriptor @garuma I'll need a re-review as I had broken a unit test that was looking for a null category. So need to update that to get a green light.

decriptor:
@CartBlanche The resx file is reformatted again :(

CartBlanche:
@decriptor unfortunately so. It literally is just the last 4 elements that were added. Both VS4M and VS4W each do different formatting :(

PR #725, opened 2020-06-16
Update Provisionator to something more modern

CartBlanche:


decriptor:
Do we need to be careful of vsmac since it still targets older versions of xcode?

CartBlanche:
@decriptor agreed. Though right now we aren't using any APIs from 11.5.

PR #738, opened 2020-06-23
Adds a new mechanism to open Modal dialogs over VS4Mac context

netonjm:
This issue was hitting some problems. Use NSApplication.SharedApplication.RunModalForWindow (window) over a GTK application has very odd behaviours with our hacks. In this case, it sets null the current NSApplication.SharedApplciation.MainWindow, and that's the reason because the focused ring is not visible and not receiving the keyevents.

Additional to this, the IDE has a default behaviour of focus the Document selected when it gets the focus, and this behaviour it was executed after the Proppy call. That's because now we wait until SynchronizationContext to execute our proppy focus call.

v6G5ysrIob

The fix doesn't breaks the Standalone

yh08ZMuOVJ

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1002094

### DIFF
--- a/Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
+++ b/Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
@@ -41,7 +41,7 @@ namespace Xamarin.PropertyEditing.Mac
 			var result = (NSModalResponse)(int)NSApplication.SharedApplication.RunModalForWindow (window);
 
 			//after run modal our FocusedWindow is null, we set the parent again
-			parentWindow.MakeKeyAndOrderFront (parentWindow);
+			parentWindow?.MakeKeyAndOrderFront (parentWindow);
 
 			System.Threading.Tasks.Task.Delay (defaultDelayTime).ContinueWith (t => {
 				responseHandler?.Invoke (result);
--- a/Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
+++ b/Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
@@ -1,0 +1,51 @@
+//
+// CocoaHelpers.cs
+//
+// Author:
+//       jmedrano <josmed@microsoft.com>
+//
+// Copyright (c) 2020 
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public static class CocoaHelpers
+	{
+		public static void RunModalForWindow (NSWindow window, NSWindow parentWindow = null, Action<NSModalResponse> responseHandler = null, int defaultDelayTime = 100)
+		{
+			//HACK: Because VS4Mac is a GTK application try force set to NSApplication.SharedApplication.RunModalForWindow 
+			//breaks the current focused window. Try only focus the ID is not enought, because our IDE on get focus (gtk) will override the current
+			//focused element, then launch a task to allow the IDE to get the focus and wait for synchcontext to focus the correct view.
+			if (parentWindow == null)
+				parentWindow = NSApplication.SharedApplication.KeyWindow;
+
+			var result = (NSModalResponse)(int)NSApplication.SharedApplication.RunModalForWindow (window);
+
+			//after run modal our FocusedWindow is null, we set the parent again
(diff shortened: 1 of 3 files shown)


garuma:
🤯

Xamarin.PropertyEditing.Mac/Controls/CollectionInlineEditorControl.cs
```diff
@@ -25,8 +25,19 @@ public CollectionInlineEditorControl (IHostResourceProvider hostResources)
 			};
 
 			this.openCollection.Activated += (o, e) => {
-				CollectionEditorWindow.EditCollection (EffectiveAppearance, HostResources, ViewModel);
-				Window.MakeFirstResponder (this.openCollection);
+				NSWindow parentWindow = NSApplication.SharedApplication.KeyWindow;
```
CartBlanche, line 28: This may be a simpler implementation of this work around/hack. Then it doesn't need the additional helper function etc.

```suggestion
				var parentWindow = NSApplication.SharedApplication.KeyWindow;

				CollectionEditorWindow.EditCollection (EffectiveAppearance, HostResources, ViewModel);

				//HACK: after a modal display the GTK FocusedWindow, VS4M is null, we set the parent again so it behaves correctly
				//TODO: Remove this Hack, once VS4M is a proper Full Cream Cocoa app.
				parentWindow?.MakeKeyAndOrderFront (parentWindow);

				System.Threading.Tasks.Task.Delay (100).ContinueWith (t => {
					parentWindow?.MakeFirstResponder (this.openCollection);
				}, System.Threading.Tasks.TaskScheduler.FromCurrentSynchronizationContext ());
```

Xamarin.PropertyEditing.Mac/Controls/CollectionInlineEditorControl.cs
```diff
@@ -25,8 +25,19 @@ public CollectionInlineEditorControl (IHostResourceProvider hostResources)
...
+				var w = new CollectionEditorWindow (hostResources, ViewModel) {
+					Appearance = EffectiveAppearance
+				};
+
+				CocoaHelpers.RunModalForWindow (w, parentWindow, responseHandler: result => {
+					if (result != NSModalResponse.OK)
+						ViewModel.CancelCommand.Execute (null);
+					else
+						ViewModel.CommitCommand.Execute (null);
+					parentWindow?.MakeFirstResponder (this.openCollection);
```
CartBlanche, line 39: Remove this line here, and instead pass in the `NSView` that we want to refocus. I'll add the suggestion there. This might make it a bit more usable too.

Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+			if (parentWindow == null)
+				parentWindow = NSApplication.SharedApplication.KeyWindow;
+
+			var result = (NSModalResponse)(int)NSApplication.SharedApplication.RunModalForWindow (window);
+
+			//after run modal our FocusedWindow is null, we set the parent again
+			parentWindow?.MakeKeyAndOrderFront (parentWindow);
+
+			System.Threading.Tasks.Task.Delay (defaultDelayTime).ContinueWith (t => {
+				responseHandler?.Invoke (result);
```
CartBlanche, line 47: add 
`parentWindow?.MakeFirstResponder (controlToFocusWhenWindowClosed);`

After the call.

Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace Xamarin.PropertyEditing.Mac
+{
+	public static class CocoaHelpers
+	{
+		public static void RunModalForWindow (NSWindow window, NSWindow parentWindow = null, Action<NSModalResponse> responseHandler = null, int defaultDelayTime = 100)
+		{
+			//HACK: Because VS4Mac is a GTK application try force set to NSApplication.SharedApplication.RunModalForWindow 
+			//breaks the current focused window. Try only focus the ID is not enought, because our IDE on get focus (gtk) will override the current
+			//focused element, then launch a task to allow the IDE to get the focus and wait for synchcontext to focus the correct view.
+			if (parentWindow == null)
```
CartBlanche, line 38: No need for this. Just grab the parent before we show the window modally.

Xamarin.PropertyEditing.Mac/CocoaHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using AppKit;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public static class CocoaHelpers
+	{
+		public static void RunModalForWindow (NSWindow window, NSWindow parentWindow = null, Action<NSModalResponse> responseHandler = null, int defaultDelayTime = 100)
```
CartBlanche, line 33:
```suggestion
		public static void RunModalForWindow (NSWindow window, NSView controlToFocusWhenWindowClosed, Action<NSModalResponse> responseHandler = null, int defaultDelayTime = 100)
```

Xamarin.PropertyEditing.Mac/Controls/CollectionInlineEditorControl.cs
```diff
@@ -25,8 +25,19 @@ public CollectionInlineEditorControl (IHostResourceProvider hostResources)
...
 			this.openCollection.Activated += (o, e) => {
-				CollectionEditorWindow.EditCollection (EffectiveAppearance, HostResources, ViewModel);
-				Window.MakeFirstResponder (this.openCollection);
+				NSWindow parentWindow = NSApplication.SharedApplication.KeyWindow;
+
+				var w = new CollectionEditorWindow (hostResources, ViewModel) {
+					Appearance = EffectiveAppearance
+				};
+
+				CocoaHelpers.RunModalForWindow (w, parentWindow, responseHandler: result => {
```
CartBlanche, line 34:
```suggestion
				CocoaHelpers.RunModalForWindow (w, this.openCollection, responseHandler: result => {
```

Xamarin.PropertyEditing.Mac/Controls/CollectionInlineEditorControl.cs
CartBlanche, line 28: This call, wouldn't be needed either.

CartBlanche:
@netonjm @garuma Hope the above suggested changes make sense. If not I can submit a commit. Let me know either way.

CartBlanche:
@netonjm Let me know if the new commit is ok and works for you too?

netonjm:
sounds good

CartBlanche:
@netonjm I tested in Proppy and in VsMac and appears to work. But just to be safe it would be nice to have someone else play with it. I'm a terrible tester :p

PR #753, opened 2020-07-25
Add Initial Support for PoliCheck and CredScan

CartBlanche:


azure-pipelines.yaml
```diff
@@ -112,3 +121,17 @@ stages:
               pr_number: $(pull_request_number)
               merge_method: 'merge'
             displayName: Merge PR# $(System.PullRequest.PullRequestNumber)
+
+  - stage: PoliCheck
```
mcumming, line 125: For consistency with what I have for the designer repo, call the stage code_analysis and a display name of "Code Analysis" and move the stage to the top of the file.

azure-pipelines.yaml
```diff
@@ -112,3 +121,17 @@ stages:
...
+    dependsOn: ['Windows', 'Mac']
```
mcumming, line 126: This is not needed and will just extend the time to completion

azure-pipelines.yaml
```diff
@@ -112,3 +121,17 @@ stages:
...
+    displayName: PoliCheck
+    jobs:
+    # Check - "PoliCheck "
+    - job: run_poli_check
```
mcumming, line 130: for consistency, use run_polichack_compliance with a display name of "Policheck Compliance"

azure-pipelines.yaml
CartBlanche: Done.

azure-pipelines.yaml
CartBlanche: Done.

azure-pipelines.yaml
CartBlanche: Done.

CartBlanche:
@mcumming Added your suggestions, including CredScan.

PR #756, opened 2020-09-08
Refactor to have individual Date and Time Controls for better mapping…

CartBlanche:
… for .Forms

This PR gets things working in the respective Stand-Alones. Separate PR will be needed for uitools support.

**Windows**
Screenshot 2020-09-29 at 16 25 49

Screenshot 2020-09-29 at 16 33 29

**MacOS**

Screenshot 2020-09-29 at 16 28 40

Screenshot 2020-09-29 at 16 28 58

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/BaseDateTimeEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/BaseDateTimeEditorControl.cs
@@ -1,0 +1,58 @@
+using System;
+using AppKit;
+using Xamarin.PropertyEditing.ViewModels;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal abstract class BaseDateTimeEditorControl<T> : PropertyEditorControl<PropertyViewModel<T>>
+	{
+		public NSDatePicker DatePicker { get; }
+
+		public BaseDateTimeEditorControl (IHostResourceProvider hostResources)
+			: base (hostResources)
+		{
+			DatePicker = new NSDatePicker {
+				ControlSize = NSControlSize.Small,
+				DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond | NSDatePickerElementFlags.YearMonthDateDay,
+				DatePickerStyle = NSDatePickerStyle.TextFieldAndStepper,
+				Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Small)),
+				TimeZone = Foundation.NSTimeZone.FromAbbreviation ("UTC"),
+				TranslatesAutoresizingMaskIntoConstraints = false,
+			};
+
+			// update the value on keypress
+			DatePicker.Activated += Editor_Activated;
+
+			AddSubview (DatePicker);
+
+			AddConstraints (new[] {
+				NSLayoutConstraint.Create (DatePicker, NSLayoutAttribute.CenterY, NSLayoutRelation.Equal, this,  NSLayoutAttribute.CenterY, 1f, 0f),
+				NSLayoutConstraint.Create (DatePicker, NSLayoutAttribute.Leading, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Leading, 1f, 0f),
+				NSLayoutConstraint.Create (DatePicker, NSLayoutAttribute.Width, NSLayoutRelation.Equal, this,  NSLayoutAttribute.Width, 1f, 0),
+				NSLayoutConstraint.Create (DatePicker, NSLayoutAttribute.Height, NSLayoutRelation.Equal, this, NSLayoutAttribute.Height, 1f,  -6f),
+			});
+		}
+
+		protected abstract void Editor_Activated (object sender, EventArgs e);
+
+		public override NSView FirstKeyView => DatePicker;
+		public override NSView LastKeyView => DatePicker;
+
+		protected override void SetEnabled ()
+		{
+			DatePicker.Enabled = ViewModel.Property.CanWrite;
+		}
+
+		protected override void UpdateAccessibilityValues ()
+		{
+			DatePicker.AccessibilityTitle = string.Format (Properties.Resources.AccessibilityDateTime, ViewModel.Property.Name);
+			DatePicker.Enabled = DatePicker.Enabled;
+		}
+
+		protected override void Dispose (bool disposing)
+		{
+			DatePicker.Activated -= Editor_Activated;
+			base.Dispose (disposing);
+		}
+	}
+}
--- a/Xamarin.PropertyEditing.Mac/Controls/DateEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/DateEditorControl.cs
@@ -1,0 +1,27 @@
+using System;
+using AppKit;
+using Xamarin.PropertyEditing.Common;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	internal class DateEditorControl : BaseDateTimeEditorControl<Date>
+	{
+
+		public DateEditorControl (IHostResourceProvider hostResources)
+			: base (hostResources)
+		{
+			DatePicker.DatePickerElements = NSDatePickerElementFlags.YearMonthDateDay;
+		}
+
+		protected override void Editor_Activated (object sender, EventArgs e)
+		{
+			ViewModel.Value = new Date (DatePicker.DateValue.ToDateTime ());
+		}
+
+		protected override void UpdateValue ()
+		{
+			if (ViewModel.Value != null)
+				DatePicker.DateValue = ViewModel.Value.DateTime.ToNSDate ();
+		}
+	}
+}
--- a/Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
@@ -1,65 +1,22 @@
 using System;
-using AppKit;
-using Xamarin.PropertyEditing.ViewModels;
 
 namespace Xamarin.PropertyEditing.Mac
 {
-	internal class DateTimeEditorControl : PropertyEditorControl<PropertyViewModel<DateTime>>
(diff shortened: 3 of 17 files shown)


netonjm (dismissed):
LGTM

garuma:
Nice, definitely going in the right direction

netonjm:
Absolutely agree @CartBlanche 👍

Xamarin.PropertyEditing.Windows/EditorPropertySelector.cs
```diff
@@ -113,7 +114,11 @@ private bool TryGetTemplate (Type type, out DataTemplate template)
 			{ typeof(TypePropertyViewModel), typeof(TypeEditorControl) },
 			{ typeof(CollectionPropertyViewModel), typeof(CollectionEditor) },
 			{ typeof(RatioViewModel), typeof(RatioEditorControl) },
-			{ typeof(AutoResizingPropertyViewModel), typeof(AutoResizingMaskEditorControl) }
+			{ typeof(AutoResizingPropertyViewModel), typeof(AutoResizingMaskEditorControl) },
+			{ typeof(PropertyViewModel<char>), typeof(StringEditorControl) },
```
CartBlanche, line 118: These 2 are added to show "an Editor" and just using the Default Editors This may need fleshing out in future.

Xamarin.PropertyEditing.Mac/Controls/DateEditorControl.cs
```diff
@@ -0,0 +1,27 @@
...
+
+		protected override void Editor_Activated (object sender, EventArgs e)
+		{
+			ViewModel.Value = new Date (DatePicker.DateValue.ToDateTime ());
+		}
+
+		protected override void UpdateValue ()
+		{
+			if (ViewModel.Value != null)
+				DatePicker.DateValue = ViewModel.Value.DateTime.ToNSDate ();
```
netonjm, line 24: It would be great add some try/catch block here.

Xamarin.PropertyEditing.Mac/Controls/DateEditorControl.cs
```diff
@@ -0,0 +1,27 @@
...
+
+		public DateEditorControl (IHostResourceProvider hostResources)
+			: base (hostResources)
+		{
+			DatePicker.DatePickerElements = NSDatePickerElementFlags.YearMonthDateDay;
+		}
+
+		protected override void Editor_Activated (object sender, EventArgs e)
+		{
+			ViewModel.Value = new Date (DatePicker.DateValue.ToDateTime ());
```
netonjm, line 18: It would be great add some try/catch block here.

Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
```diff
@@ -1,65 +1,22 @@
...
+		protected override void Editor_Activated (object sender, EventArgs e)
 		{
-			this.datePicker.Enabled = ViewModel.Property.CanWrite;
-		}
-
-		protected override void UpdateAccessibilityValues ()
-		{
-			this.datePicker.AccessibilityTitle = string.Format (Properties.Resources.AccessibilityDateTime, ViewModel.Property.Name);
-			this.datePicker.Enabled = this.datePicker.Enabled;
+			ViewModel.Value = DatePicker.DateValue.ToDateTime ();
```
netonjm, line 14: It would be great add some try/catch block here.

Xamarin.PropertyEditing.Mac/Controls/DateTimeEditorControl.cs
```diff
@@ -1,65 +1,22 @@
...
 		}
 
-		protected override void Dispose (bool disposing)
+		protected override void UpdateValue ()
 		{
-			this.datePicker.Activated -= Editor_Activated;
-			base.Dispose (disposing);
+			DatePicker.DateValue = ViewModel.Value.ToNSDate ();
```
netonjm, line 19: It would be great add some try/catch block here.

Xamarin.PropertyEditing.Mac/Controls/TimeEditorControl.cs
```diff
@@ -0,0 +1,27 @@
...
+	{
+		public TimeEditorControl (IHostResourceProvider hostResources)
+			: base (hostResources)
+		{
+			DatePicker.DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond;
+		}
+
+		protected override void Editor_Activated (object sender, EventArgs e)
+		{
+			ViewModel.Value = new Time (DatePicker.DateValue.ToDateTime ());
```
netonjm, line 17: It would be great add some try/catch block here.

Xamarin.PropertyEditing.Mac/Controls/TimeEditorControl.cs
```diff
@@ -0,0 +1,27 @@
...
+		}
+
+
+		protected override void UpdateValue ()
+		{
+			if (ViewModel.Value != null)
+				DatePicker.DateValue = ViewModel.Value.DateTime.ToNSDate ();
```
netonjm, line 24: It would be great add some try/catch block here.

Xamarin.PropertyEditing.Windows/EditorPropertySelector.cs
netonjm: allow editor providers would be awesome

CartBlanche:
@netonjm Thanks for the `try...catch` suggestions, my only concern is our other editors don't make extensive use of them, so feel if we do that we should be consistent, possibly in a separate specific PR, and add them across the board. Also keep in mind that adding `try...catch` has extra performance overheads. So we may need to profile that and compare with and without.

Xamarin.PropertyEditing.Mac/Controls/BaseDateTimeEditorControl.cs
```diff
@@ -0,0 +1,58 @@
...
+
+		public BaseDateTimeEditorControl (IHostResourceProvider hostResources)
+			: base (hostResources)
+		{
+			DatePicker = new NSDatePicker {
+				ControlSize = NSControlSize.Small,
+				DatePickerElements = NSDatePickerElementFlags.HourMinuteSecond | NSDatePickerElementFlags.YearMonthDateDay,
+				DatePickerStyle = NSDatePickerStyle.TextFieldAndStepper,
+				Font = NSFont.SystemFontOfSize (NSFont.SystemFontSizeForControlSize (NSControlSize.Small)),
+				TimeZone = Foundation.NSTimeZone.FromAbbreviation ("UTC"),
```
joshspicer, line 19: Is timezone something that needs to be specified?

Xamarin.PropertyEditing.Windows/DateToTextConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Windows.Markup;
+using Xamarin.PropertyEditing.Common;
+
+namespace Xamarin.PropertyEditing.Windows
+{
+	[ValueConversion (typeof (Date), typeof (string))]
+	internal class DateToTextConverter : MarkupExtension, IValueConverter
+	{ 
+		public object Convert (object value, Type targetType, object parameter, CultureInfo culture)
+			=> !(value is Date dateValue) ? DependencyProperty.UnsetValue
```
joshspicer, line 14: The formatting is a little funny to me.  Maybe just drop the `DependencyProperty.UnsetValue` down a line?

Xamarin.PropertyEditing/Common/Date.cs
```diff
@@ -0,0 +1,60 @@
...
+		public bool Equals (Date other)
+		{
+			if (other == null)
+				return false;
+			return this.dateTime.Equals (other);
+		}
+
+		public override int GetHashCode ()
+		{
+			var hashCode = 1861433795;
```
joshspicer, line 39: What's this number - `1861433795` ?

joshspicer (approved):

Xamarin.PropertyEditing.Mac/Controls/DateEditorControl.cs
mcumming: Actually, I don't agree here, `DatePicker.DateValue` should always return a value, but I do not know what the behavior of `ToDateTime()`, I am making an assumption that it will always return a valid value for the `Date` ctor, if that is not true, then we should be validating that value before constructing the `Date`.

The Event processor for `Editor_Activated` should be handling any exceptions thrown by event handlers so that all event handlers are called properly.

mcumming (approved):

Xamarin.PropertyEditing.Mac/Controls/BaseDateTimeEditorControl.cs
netonjm: yes, all times needs to be converted to UTC

Xamarin.PropertyEditing/Common/Date.cs
netonjm: probably this could be added as a const

netonjm:
> @netonjm Thanks for the `try...catch` suggestions, my only concern is our other editors don't make extensive use of them, so feel if we do that we should be consistent, possibly in a separate specific PR, and add them across the board. Also keep in mind that adding `try...catch` has extra performance overheads. So we may need to profile that and compare with and without.

Any possibility of catching an exception that could occur (as is the case in case of set a DateTime) should be ahead of the little memory savings that we can have.

CartBlanche:
> Any possibility of catching an exception that could occur over runtime (as is the case in case of set a DateTime) should be ahead of the little memory savings that may cause.

When I mention performance, I wasn't thinking about memory, but more speed. The setting `DateTime` incorrectly scenario should be handled. The current Xamarin.Forms `DatePicker` element has already been using `DateTime` for over year, maybe 2 by now. I've not seen any DevDiv issue relating to it. Have you?

CartBlanche:
@mcumming the `ToDateTime` extension method is here - https://github.com/xamarin/Xamarin.PropertyEditing/blob/main/Xamarin.PropertyEditing.Mac/Controls/DateExtensions.cs#L33-L41

PR #763, opened 2021-03-16
Don't allow property editor panel to get the focus

BretJohnson:
Hitting tab should never set focus to the property editor panel itself. Fixed [AB#1251446](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1251446)


BretJohnson:
@gitbot backport d16-9

Aguilex (approved):

PR #765, opened 2021-03-25
Migrate to OneLocBuild

mcumming:


.build/automation/stages/localization-handback.yml
```diff
@@ -0,0 +1,85 @@
...
+              #-- Import Invoke-Git Module function
+              Import-Module $(Build.SourcesDirectory)\.build\automation\scripts\Invoke-Git.psm1 -Force -DisableNameChecking
+              #--Clear Stage
+              Invoke-Git reset HEAD -- .
+              #-- Create new branch
+              $branchName = "$(LocBranchPrefix)/$(new-guid)"
+              Invoke-Git checkout -B ${branchName}
+              Write-Host ("##vso[task.setvariable variable=HANDBACK_BRANCH_NAME;]$branchName")
+              #-- Stage Build Changes
+              Invoke-Git add Designer
```
mauroa, line 51: I don't think this would work since those folders are not on the source repo, right?

.build/automation/stages/localization-handback.yml
```diff
@@ -0,0 +1,85 @@
...
+                title = "[LOC_HB] checkin of localized string files"
+                head = "$(HANDBACK_BRANCH_NAME)"
+                base = "main"
+                maintainer_can_modify = $true
+              }
+
+              $headers = @{ Authorization = "token $(GitHub.Token)" }
+
+              # let it throw
+              $response = Invoke-WebRequest -UseBasicParsing -Method POST -Headers $headers -Uri "https://api.github.com/repos/Xamarin/UITools/pulls" -Body ($payload | ConvertTo-json)
```
mauroa, line 75: Same for this, UITools is invalid

.build/automation/variables.yml
```diff
@@ -0,0 +1,29 @@
...
+- name: isMainBranch
+  value: $[eq(variables['Build.SourceBranch'], 'refs/heads/main')]
+- name: isLocBranch
+  value: $[eq(variables['Build.SourceBranch'], 'refs/heads/loc')]
+- name: isTargetMainBranch
+  value: $[eq(variables['System.PullRequest.TargetBranch'], 'refs/heads/main')]
+- name: isTargetLocBranch
+  value: $[eq(variables['System.PullRequest.TargetBranch'], 'refs/heads/loc')]
+- name: isLocPRBranch
+  value: $[startsWith(variables['System.PullRequest.SourceBranch'], 'refs/heads/$(LocBranchPrefix)')]
```
mauroa, line 19: I tried using `$(...)` to reference variables from a new variable declaration and it didn't work. Looks like you should use `variables['...']`, but it maybe worth double checking

.config/LocProject.json
```diff
@@ -0,0 +1,17 @@
+{
+    "Projects": [
+      {
+        "LanguageSet": "VS_Main_Languages",
+        "LocItems": [
+            {
+                "SourceFile": "Xamarin.PropertyEditing\Resources\Resources.resx",
```
mauroa, line 7: You are missing to escape the `SourceFile` path, aren't you?

mauroa (changes requested):

.build/automation/variables.yml
mcumming: This working correctly in UITools, so I am going to leave it as is.

rmarinho: didn't work for me either

decriptor (approved):

PR #766, opened 2021-04-08
Bump provisioned dependencies

BretJohnson:
Bump provisioned dependencies to match what's currently used
for Designer, getting the latest Xamarin.Mac.

This is needed because the Xamarin.Mac.dll path changed and we
need to update the csproj reference to it accordingly, also done here.

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Xamarin.PropertyEditing.Mac.csproj
+++ b/Xamarin.PropertyEditing.Mac/Xamarin.PropertyEditing.Mac.csproj
@@ -10,7 +10,7 @@
     <Reference Include="System.Core" />
     <Reference Include="System.Drawing" />
     <Reference Include="Xamarin.Mac">
-      <HintPath>/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/lib/x86_64/full/Xamarin.Mac.dll</HintPath>
+      <HintPath>/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/lib/64bits/full/Xamarin.Mac.dll</HintPath>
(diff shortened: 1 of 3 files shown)


decriptor:
I'm not sure why this built and worked, but things like this have to be changed:
https://github.com/xamarin/Xamarin.PropertyEditing/blob/1fd6c0b7f1221222bc93d0d509e267c64e19cb45/Xamarin.PropertyEditing.Mac/Xamarin.PropertyEditing.Mac.csproj

- <MacFile Include="/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/lib/x86_64/full/Xamarin.Mac.dll"/>
+ <MacFile Include="/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/lib/64bits/full/Xamarin.Mac.dll"/>

decriptor (changes requested):
See my other comment about the XamMac path

BretJohnson:
@decriptor Yeah, I meant to update that. Thx. Fixed.

decriptor:
@BretJohnson no worries. I've been dealing with it a bunch lately so it is on the top of my mind :) The other slightly older change was the location of libxammac.dylib, but I don't think it is referenced in here.

BretJohnson:
@mcumming And now the build fails, since the agent's installed VSMac doesn't have the latest Xamarin.Mac and we don't provision. Should we update to use different YAML, that does provisioning? Can you make that change?

BretJohnson:
@mcumming - ping on this

decriptor:
Closing in favor of #767

PR #770, opened 2021-06-29
Fix edit box focus border contrast

BretJohnson:
Fix [AB#1251444](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1251444). The rule is that the border for edit boxes (and all controls) when focused must have at least a 3:1 contrast difference with the border when unfocused - there must enough contrast difference to make obvious to all that the visual state changed. That wasn't the case for our edit boxes before but is now. Updated to use the same themed border & text colors as the as the Cider colors, used by the WPF designer, which we've kept consistent with in the past.

Checkboxes and buttons already show a dashed outline for a focus rect. That's not the current standard for VS UI controls, but it's what it's always done and is fine from an a11y point of view, since there's even more obvious contrast, so that didn't change here. It was only text boxes that had an a11y problem.


mcumming:
Is there a similar issue on the Mac side?

PR #771, opened 2021-07-30
Fixes double disclosure triangle in Property panel

netonjm:
Fixes VSTS #1340425: Property pane has double chevrons

old

image

new

CleanShot 2021-07-30 at 12 11 04@2x

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyList.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyList.cs
@@ -119,6 +119,8 @@ namespace Xamarin.PropertyEditing.Mac
 				return true;
 			}
 
+			public override CGRect FrameOfOutlineCellAtRow (nint row) => CGRect.Empty;
+
 			public override bool BecomeFirstResponder ()
 			{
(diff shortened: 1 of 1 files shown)


mrward (approved):

netonjm:
cc @BretJohnson

Xamarin.PropertyEditing.Mac/PropertyList.cs
```diff
@@ -119,6 +119,8 @@ public override bool ValidateProposedFirstResponder (NSResponder responder, NSEv
 				return true;
 			}
 
+			public override CGRect FrameOfOutlineCellAtRow (nint row) => CGRect.Empty;
```
Therzok, line 122: Why do we have a custom disclosure triangle?

netonjm: Because the current logic expects this item to handle the managed NextKeyResponder, so I wouldn't change this to avoid break things.

BretJohnson (approved):

PR #785, opened 2022-01-06
Switch Mac project to .net6

BretJohnson:


BretJohnson:
This PR branch is referenced from https://github.com/xamarin/UITools/pull/608

Therzok (approved):

drasticactions (approved):

.build/automation/stages/validate.yml
```diff
@@ -12,6 +12,19 @@ stages:
             clean: true
             persistCredentials: true
 
+          - task: CmdLine@1
```
mcumming, line 15: Why is this required? You are essentially setting the value back to itself?

mcumming (changes requested):

.build/automation/stages/validate.yml
BretJohnson: You are asking about the Set Token step? I'm not sure why (or if) it's needed, but we seem to do everywhere before calling the provisionator. I copied it from the Windows section above. It's also in Hot Reload here: https://github.com/xamarin/UITools/blob/a90b49522b5aadc38c8f63992cf80c1e63d60e21/HotReload/build/automation/stages/validate.yml#L189
Do think it doesn't actually do anything? It seems like it's setting the GITHUB_TOKEN environment variable from the pipeline property.

BretJohnson: @mcumming ^^^

BretJohnson: Indeed, this wasn't needed. I removed it, from both Mac and Windows.

mcumming (approved):

Therzok:
Have we tried inserting this in VSMac?

PR #794, opened 2022-02-25
Add cast to fix bug with numeric conversion

BretJohnson:
With the macos P13 API type changes, an issue was introduced since
casting an ObjCRuntime.nfloat to an nint (aka IntPtr) doesn't work
properly - (nint)(ObjCRuntime.nfloat)23 for instance will evaluate
to 0x4037000000000000, the floating point bits are just reinterpreted
as an int rather than actually being converted.

That in turn caused huge sizes for width and height causing
new CGBitmapContext to throw an exception.

Adding the double cast here fixes this, for now.

Soon ObjCRuntime.nfloat will go away, replaced by
System.Runtime.InteropServices.NFloat, but for now
this workaround makes us work.

Fixes:
[AB#1487089](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1487089)
[AB#1487070](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1487070)
[AB#1487089](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1487089)

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/CommonBrushLayer.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/CommonBrushLayer.cs
@@ -84,8 +84,11 @@ namespace Xamarin.PropertyEditing.Mac
 		public NSImage RenderPreview ()
 		{
 			var scale = this.ContentsScale;
-			nint h = (nint)(this.Bounds.Height * scale);
-			nint w = (nint)(this.Bounds.Width * scale);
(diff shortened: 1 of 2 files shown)


sandyarmstrong (approved):

BretJohnson:
@gitbot backport main

PR #797, opened 2022-03-04
Enable MicroBuild Signing

thsparks:
This enables MicroBuild signing of assemblies from this repo, which is used as a submodule in UITools (where the actual signing pipeline will be run).

For normal builds (outside of the signing pipeline), MicroBuild will not be setup, so these steps will just be no-ops.


mcumming:
We may not have to do this, by using a Directory.Builds.props in the repo that submodules this we can add the required details to sign the assemblies created by this repo

mrward (approved):

Xamarin.PropertyEditing/Xamarin.PropertyEditing.csproj
```diff
@@ -20,4 +26,15 @@
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
   </ItemGroup>
+
+  <Target Name="GetFilesToSign" BeforeTargets="SignFiles" Condition=" '$(Configuration)' == 'Release' Or '$(Configuration)' == 'ReleaseWindows'">
```
sandyarmstrong, line 30: Can you explain why this and the other `FileToSign` `ItemGroup` are needed? I thought MicroBuild automatically identified these assemblies. Are you using a different signing certificate or something?

sandyarmstrong: `Microsoft400` seems to be the standard cert so I'm not sure why this is needed. You can enable or disable based on configuration in your pipeline.

Xamarin.PropertyEditing.Windows/Xamarin.PropertyEditing.Windows.csproj
```diff
@@ -30,6 +30,11 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core">
```
sandyarmstrong, line 34: I agree that you can bring these in via `Directory.Build.targets`, whether in this repo or in the repo that submodule this.

Xamarin.PropertyEditing/Xamarin.PropertyEditing.csproj
thsparks: My understanding was each dll inside the vsix needed to be called out in this way. Mostly going off of samples and guidance provided through out wiki.


@mattnorflus , do we not actually need to do it this way?

sandyarmstrong: I think that's only for nested content. For this csproj's assembly, it should be signed automatically by the microbuild targets. But @mattnorflus should know better than I do.

sandyarmstrong: https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/27099/Guidance-for-Xamarin-MicroBuild-Signing-Migration?anchor=handling-nested-content

mattnorflus: This is correct, the `FilesToSign` `ItemGroup` is how MicroBuild knows which files to sign in the SignFiles target. Each file that ends up in the vsix must be signed in this way prior to getting packaged into the vsix.

sandyarmstrong: And then there's also [this guidance](https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/27099/Guidance-for-Xamarin-MicroBuild-Signing-Migration?anchor=avoid-oversigning-if-building-multiple-configurations) for avoiding signing during non-Release builds, but I feel like that is more easily solved by not installing the signing pluging for non-Release builds in your pipeline.

mattnorflus: For additional clarification, the [core MicroBuild guidance](https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/650/MicroBuild-Signing?anchor=modify-your-project-file-to-include-%60filestosign%60-itemgroup) states that the `FilesToSign` `ItemGroup` is required for all files that need to be sign. "In order to add signing to your projects, modify your project files to add the following item group"

sandyarmstrong: Hmm, interesting, maybe it's `Microsoft.VisualStudio.Internal.MicroBuild.VisualStudio` that automatically populates `FilesToSign`? We definitely don't set it ourselves in the editor build.

thsparks: So if I'm understanding correctly, it sounds like this is expected, at least for the version of MicroBuild we're using.

thsparks:
Unfortunately, I don't think adding it in the UITools externals\Directory.Builds.props works, since this project's own Directory.Build.props simply blocks it from taking effect. I did try to leverage the Directory.Build.props in here, though... LMK if I'm missing something.

---
In reply to: [1059187405](https://github.com/xamarin/Xamarin.PropertyEditing/pull/797#issuecomment-1059187405) [](http://example.com/codeflow?ancestors=1059187405,1059187405)

Xamarin.PropertyEditing/Xamarin.PropertyEditing.csproj
sandyarmstrong: I guess so. It seems the internal package does a bit more heavy lifting automatically.

mcumming (approved):

PR #800, opened 2022-03-17
Use PreviewKeyDown for Ctrl+Space to show menu

BretJohnson:
Ctrl+Space is the keyboard short that can
be used to show the property editor menu
for a property (with options like Reset,
Convert to Local Value, etc.). Previously, that
was detected via a KeyBinding in XAML, which
worked for text boxes but not controls like
checkbox and button, which handle the
key down event themselves. Now we
use PreviewKeyDown instead, handling that
in C# code, to work for all control
types.

Fixes [AB#1491229](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1491229)

### DIFF
--- a/Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
+++ b/Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
@@ -117,6 +117,7 @@ namespace Xamarin.PropertyEditing.Windows
 			this.propertyButton = (PropertyButton) GetTemplateChild ("propertyButton");
 
 			this.propertyContainer = (Border) GetTemplateChild ("propertyContainer");
+			// Since the template never changes, the handler only gets added once and there's no need to unsubscribe
 			this.propertyContainer.AddHandler (Border.PreviewKeyDownEvent, new KeyEventHandler (PropertyContainer_PreviewKeyDown));
 		}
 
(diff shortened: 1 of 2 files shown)


Xamarin.PropertyEditing.Windows/PropertyPresenter.cs
```diff
@@ -112,6 +113,11 @@ public override void OnApplyTemplate ()
 
 			this.variationRow = (RowDefinition) GetTemplateChild ("variationRow");
 			this.variationsList = GetTemplateChild ("variationsList") as ItemsControl;
+
+			this.propertyButton = (PropertyButton) GetTemplateChild ("propertyButton");
+
+			this.propertyContainer = (Border) GetTemplateChild ("propertyContainer");
+			this.propertyContainer.AddHandler (Border.PreviewKeyDownEvent, new KeyEventHandler (PropertyContainer_PreviewKeyDown));
```
netonjm, line 120: there is any mechanism to unsuscribe it?

BretJohnson: Since the template never changes, OnApplyTemplate is only called once and there shouldn't be a need to call RemoveHandler. But I added a comment, to make that clearer. BrushEditorControl has the same pattern.

spadapet (approved):

PR #806, opened 2022-03-21
Backport MicroBuild signing changes to d17-0

thsparks:
With the move away from the "sign-from-github-esrp" ("Groovy") pipeline, we need to shift our assembly signing to happen alongside the build itself, which is more secure. Without these changes, we will not be able to push new, signed assemblies from this branch.


BretJohnson (approved):

BretJohnson:
LGTM. Are you planning to take these changes to the d17-1 branch as well?

thsparks:
> LGTM. Are you planning to take these changes to the d17-1 branch as well?

TBD, would like to get input from the rest of the team on this, but focusing on 17-0 in the meantime while that discussion takes place.

PR #812, opened 2022-04-19
Remove UpdateExpansions

BretJohnson:
Remove the call to UpdateExpansions after ReloadData in response to
OnPropertiesChanged (e.g. called when selection changes). This
doesn't seem necessary, as when views are recreated their
expansion status is set correctly originally and AppKit
remembers expansion state after.

Further, this fixes [AB#1521131](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1521131), where the Name and Type rows are
sometimes blank - apparently expanding manually, and doing it
synchronously like that, causes the NSOutlineView
to get confused and not show the top row sometimes.


Therzok (approved):

BretJohnson:
@gitbot backport main

PR #814, opened 2022-04-25
Fix Build Pipeline (MicroBuild Signing) in d16-11

thsparks:
With the move away from the "sign-from-github-esrp" ("Groovy") pipeline, we need to shift our assembly signing to happen alongside the build itself, which is more secure. Without these changes, we will not be able to push new, signed assemblies from this branch.

This change cherry-picks the following commits:
1. 6359edaa5c81a0d1697a1ad36807b74388f8ebb2 (main change)
2. 626060a19b286ac6d06f63efc31202b9e34b461b (optimization)

Test run in the UI Tools pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6067078&view=results


thsparks:
Closing this, as these changes are out-dated and should be simplified after the UI Tools signing optimizations that are now in Main.

PR #816, opened 2022-05-27
Use a variable width for the property search filter

BretJohnson:
Make VSMac have variable width for the property search
area instead of the fixed width of 150. That makes
it behave more like Windows, where the width is also
variable, allowing more space for the user to type.

It also fixes [AB#1522517](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1522517).

Old UI:
image

New UI:
image

Video of resize:
https://user-images.githubusercontent.com/245892/170719196-96fceb09-cf18-4159-8b5c-47c3f6cc2022.mov


netonjm:
could you add some screencast of how the resize works now?

Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
```diff
@@ -192,8 +192,8 @@ private void Initialize ()
 				NSLayoutConstraint.Create (this.tabStack, NSLayoutAttribute.Bottom, NSLayoutRelation.Equal, this.header, NSLayoutAttribute.Bottom, 1, 0),
 				NSLayoutConstraint.Create (this.tabStack, NSLayoutAttribute.Right, NSLayoutRelation.LessThanOrEqual, this.propertyFilter, NSLayoutAttribute.Left, 1, 0),
 
+				NSLayoutConstraint.Create (this.propertyFilter, NSLayoutAttribute.Left, NSLayoutRelation.Equal, this.tabStack, NSLayoutAttribute.Right, 1, 20),
```
netonjm, line 195: We should use Leading and Trailing, instead Left-Right

BretJohnson: updated

netonjm (approved):

Xamarin.PropertyEditing.Mac/PropertyEditorPanel.cs
netonjm: Great thanks!

Therzok (approved):

iainx:
This pad should be using the declarative toolbar system and not creating a toolbar in code manually

Therzok:
@iainx is right. I initially assumed this used NSTextFinder, but it does not:
https://blog.timschroeder.net/2012/01/12/nstextfinder-magic/
https://developer.apple.com/documentation/appkit/nstextfinder?language=objc

netonjm:
What I understand is that this toolbar would be implemented in Proppy because it is an external OpenSource library (despite the fact that we are the main users of it), and that it can be very useful to the users, to save time implementing their own. If the control provides us with API's to be able to hide and implement ours in and easy and transparent way, I don't see anything wrong with it being included in the library. But yes, it seems the search is not using NSTextFinder.

hbons:
We got updated designs for toolbars+search that @iainx and @netonjm are working on (https://www.figma.com/file/TDIvAsDxmNmqPbe2HuH1M8/Spec-%E2%80%93-Windows-and-Toolbars?node-id=120%3A3093). Can we wait until that work is done so all our tool windows are consistent?

BretJohnson:
@hbons - I updated the margins, adding 1 pixel to the top and bottom margins for the search control and reduced the left margin to 10 pixels. See below. I played around with various options here & ended up with this, which looks good to me but can you please take a look and ensure spacing is exactly right. Thanks.

image

image

image

BretJohnson:
After discussing with Hylke I removed some of the spacing to the right of the toolbox search, so now it exactly matches the space to the right of the rightmost toolbox button:

image

PR #817, opened 2022-05-27
Avoid excess margins on left/right of property editor

BretJohnson:
The property editor has noticeably wider margins on the left & right
in 17.x as compared to 8.x. It's especially noticeable on the
right hand side. That may be caused by the default NSTableView Style
changing for macOS apps linked with BigSur and
above to no longer be FullWidth,
described here: https://stackoverflow.com/questions/63175168/nstableview-has-unwanted-space-above-first-row

This change fixes that, setting the property table style to
NSTableViewStyleFullWidth, so rows fill the full width of
the table.

This fix also helps with the issues in [AB#1514437](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1514437). That
bug itself (where the scroll bar goes away) doesn't
seem to happen anymore, but I spotted the excess
margins while investigating it.

This PR also adjusts the left and right margins and the spacing between
the label and controls, as @hbons recommends below. See final screenshots in the comments.

Note that this PR doesn't contain the suggested change by @hbons to add extra
spacing at the top of the top of the scrollview, partially because I couldn't find a good
way to do that, even with Jose's help. We can make that change later in a separate PR
if we want it.


netonjm (dismissed):

Therzok (approved):

hbons:
Agree the spacing can be improved. This proposal needs a few tweaks though. I'll have a look.

hbons:
image

@BretJohnson here's some tweaks based on the screenshot of your patch.

The labels are getting a bit too close to the left edge and makes the window looks messy. By using the same 16pt padding on the left and right of the window it looks better (the left padding needs to increase, but the right padding can be reduced). Use 12pt of padding at the top until the x-height of the category label. We can also reduce the gap between label an control to 6pt. Hope that makes sense. Let me know. :)

BretJohnson:
@gitbot rebase

BretJohnson:
Here's the latest (finally) with the changes @hbons suggested (except for adding some extra margin at the top of the scrollview, as I haven't figured out how to do that yet, even with help from @netonjm). Let's at least get these changes in.

image

BretJohnson:
Merging. If we want to add more space at the top we can do that in a separate PR.

netonjm (approved):

PR #818, opened 2022-06-14
Fallback PropertyEditingResource in app bundle resources directory

netonjm:
HostResourceProvider expects the PropertyEditingResource bundle directory in the same directory than the assembly, but in VSMac we have this bundle duplicated in 2 locations. After some refactoring in the PropertyPad integration this will be only in : bundlePath = /Resources/PropertyEditingResource
So added some code to don't break and check new paths

### DIFF
--- a/Xamarin.PropertyEditing.Mac/HostResourceProvider.cs
+++ b/Xamarin.PropertyEditing.Mac/HostResourceProvider.cs
@@ -17,8 +17,14 @@ namespace Xamarin.PropertyEditing.Mac
 
 		public HostResourceProvider ()
 		{
-			var containingDir = Path.GetDirectoryName (typeof (HostResourceProvider).Assembly.Location);
-			var bundlePath = Path.Combine (containingDir, "PropertyEditingResource.bundle");
+			var bundlePath = NSBundle.MainBundle.PathForResource ("PropertyEditingResource", "bundle");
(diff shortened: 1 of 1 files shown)


Xamarin.PropertyEditing.Mac/HostResourceProvider.cs
```diff
@@ -19,6 +19,11 @@ public HostResourceProvider ()
 		{
 			var containingDir = Path.GetDirectoryName (typeof (HostResourceProvider).Assembly.Location);
 			var bundlePath = Path.Combine (containingDir, "PropertyEditingResource.bundle");
+			if (!Directory.Exists(bundlePath))
+			{
+				//if the directory is not in place we fallback into bundle resources directory
+				bundlePath = NSBundle.MainBundle.PathForResource ("PropertyEditingResource", "bundle");
```
Therzok, line 25: Which one is more likely to happen? Maybe we can invert the two.

Therzok (dismissed):

mrward (dismissed):

Xamarin.PropertyEditing.Mac/HostResourceProvider.cs
netonjm: fixed

BretJohnson (approved):

PR #820, opened 2022-07-13
Stores property list delegate to avoid GC collect it

netonjm:
Using registrar and netcore6 makes GC be more aggressive collecting objects, to ensure delegate is not collected we keep the reference.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1576420


mrward (approved):
Is there a bug associated with this change?

PR #821, opened 2022-07-21
[HistoryLayer] On viewmodel changes we ensure our viewmodel is not null and set default color

netonjm:
Fix [AB#1569130](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1569130)

Due to the issue https://github.com/xamarin/xamarin-macios/issues/15600, when a CGColor is allocated and its native object about to be passed to PInvoke when setting CALayout.BackgroundColor, it's possible for it be GC'ed, which CFReleases the CGColor, before the PInvoke completes and BackgroundColor gets a chance to CFRetain it. In that case, the BackgroundColor can point to a deleted object. Until there's a real fix for that race condition with https://github.com/xamarin/xamarin-macios/issues/15600 fixed, we workaround it here by doing a GC.KeepAlive ensuring that the object lives until the PInvoke completes.

Also, in some scenarios looks like ViewModel is not yet set in OnViewModelChanged event, which could cause a NRE. We now protect against that too.

[CleanShot 2022-07-21 at 17 38 24@2x]

(https://github.com/xamarin/Xamarin.PropertyEditing/blob/dev/netonjm/fix-1569130/Xamarin.PropertyEditing.Mac/Controls/Custom/SolidColorBrushEditorViewController.cs#L33-L47)

### DIFF
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
@@ -78,7 +78,11 @@ namespace Xamarin.PropertyEditing.Mac
 		{
 			LayoutIfNeeded ();
 			current.BackgroundColor = interaction.Color.ToCGColor ();
-			previous.BackgroundColor = interaction.ViewModel.InitialColor.ToCGColor ();
+			//there are some scenarios where viewmodel could be null
+			var initialColor = interaction.ViewModel?.InitialColor;
+			if (initialColor != null) {
+				previous.BackgroundColor = initialColor.ToCGColor ();
+			}
 			last.BackgroundColor = interaction.LastColor.ToCGColor ();
 		}
 
--- a/Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
+++ b/Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
@@ -3,6 +3,7 @@ using AppKit;
 using CoreAnimation;
 using CoreGraphics;
 using ObjCRuntime;
+using Xamarin.PropertyEditing.Drawing;
 
 namespace Xamarin.PropertyEditing.Mac
 {
@@ -77,13 +78,20 @@ namespace Xamarin.PropertyEditing.Mac
 		public override void UpdateFromModel (EditorInteraction interaction)
 		{
 			LayoutIfNeeded ();
-			current.BackgroundColor = interaction.Color.ToCGColor ();
+			SetCALayerBackgroundColor (current, interaction.Color);
 			//there are some scenarios where viewmodel could be null
-			var initialColor = interaction.ViewModel?.InitialColor;
-			if (initialColor != null) {
-				previous.BackgroundColor = initialColor.ToCGColor ();
-			}
-			last.BackgroundColor = interaction.LastColor.ToCGColor ();
+			if (interaction.ViewModel != null)
+				SetCALayerBackgroundColor (previous, interaction.ViewModel.InitialColor);
+			SetCALayerBackgroundColor (last, interaction.LastColor);
+		}
+
+		static void SetCALayerBackgroundColor(CALayer layer, CommonColor color)
+		{
+			// We need to use GC.KeepAlive to ensure that the CGColor doesn't get garbage collected (and thus the
+			// native object Released) before the BackgroundColor assignment is complete
+			CGColor cgColor = color.ToCGColor ();
+			layer.BackgroundColor = cgColor;
+			GC.KeepAlive (cgColor);
 		}
 
 		public override void UpdateFromLocation (EditorInteraction interaction, CGPoint location)


Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
```diff
@@ -78,7 +78,11 @@ public override void UpdateFromModel (EditorInteraction interaction)
 		{
 			LayoutIfNeeded ();
 			current.BackgroundColor = interaction.Color.ToCGColor ();
-			previous.BackgroundColor = interaction.ViewModel.InitialColor.ToCGColor ();
+			//there are some scenarios where viewmodel could be null
+			var initialColor = interaction.ViewModel?.InitialColor;
```
Therzok, line 82: The problem is that these CGColor instances are not stored as fields.

Therzok: You need fields for these colors, xamarin.mac only sets the native pointer here, and they could go away.

Therzok (changes requested):
Fix doesn't look right.

Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
BretJohnson: @Therzok - Are you saying that if CALayer.BackgroundColor is set to some CGColor, that alone won't keep it from getting GC'ed so you need to have another reference to that same color stored somewhere to prevent the GC? Weird - that seems like an easy mistake to make if true.

Therzok: So, there are different things at play here. NSColor used lifetime handling using NSObject wrappers with a ToggleRef. You can see that via native property annotations, like `strong`. In some cases, Xamarin.Mac is overzealous and adds `strong` managed fields for properties held `weak` by native.

CGColor uses old-school CFRetain/CFRelease lifetimes. Native holds onto a ref for the object, so something here doesn't really make sense..

> The value of this property is retained using the Core Foundation retain/release semantics. This behavior occurs despite the fact that the property declaration appears to use the default assign semantics for object retention.

```
QuartzCore!CA::AttrList::set(unsigned int, _CAValueType, void const*)
QuartzCore!CA::Layer::setter(unsigned int, _CAValueType, void const*)
QuartzCore!-[CALayer setBackgroundColor:]
VisualStudio!
Xamarin.Mac.dll!CoreAnimation.CALayer::set_BackgroundColor
Xamarin.PropertyEditing.Mac.dll!Xamarin.PropertyEditing.Mac.HistoryLay
```

Why would it crash in this code? If anything in here _were_ null, we'd crash in ToCGColor if anything was null here before setting it. So, I think the actual fix is to use something like `GC.KeepAlive` or store it in a field to avoid it getting GC'd while the PInvoke is happening.

Therzok: In short: GC ran and collected the CGColor before native retained it.

Therzok: > Weird - that seems like an easy mistake to make if true.

You're dealing with code that's almost identical to a plain PInvoke :p

BretJohnson: @Therzok - Ouch. And thanks. I would have hoped that Xamarin.Mac would have had some code to keep objects alive when they are passed along in PInvoke, but apparently not. Is there a pattern that's used in other places to address this? I was thinking I'd create a helper method to set CALayer.BackgroundColor, with the GC.KeepAlive, but if there's an existing pattern I could use that instead.

Therzok: No need to add _more_ APIs. The problem only exhibits with CoreFoundation types, not with NSObject types.

BretJohnson: @Therzok - I made an update to use GC.KeepAlive. Can you take a look at the latest change. As there are three places BackgroundColor is set, abstracting it to a separate method seemed to make sense.

Therzok (approved):

Xamarin.PropertyEditing.Mac/Controls/Custom/HistoryLayer.cs
Therzok: I opened a bug against Xamarin.Mac, so let's see if they can do anything in the code-generation layer: https://github.com/xamarin/xamarin-macios/issues/15600

Therzok (changes requested):
Requesting changes again, the commit doesn't explain _why_ this fixes it.

BretJohnson:
@Therzok I updated the PR comment to explain the issue. I'll do a squash when merging so it becomes the commit comment too.

Therzok (approved):

PR #827, opened 2023-01-12
Show proper expanded icon by default

BretJohnson:
Fix so that when a category is expanded by default the UI shows the proper icon (the expanded icon, not collapsed).

There are two main changes here:
1. Remove the code added in https://github.com/xamarin/Xamarin.PropertyEditing/commit/96561f19bf1c81e24c852e0a88ef46bd72e05e38 to support custom disclosures for Catalina. That's no longer required in newer versions of macOS. Now we just let AppKit handle setting the right disclosure icon.
2. Add call to UpdateExpansions in OnPropertiesChanged, so that the default expansion state is correct when the property editor context changes, ensuring that categories are autoexpanded when they are supposed to be.

Fixes AB#1559942

### DIFF
--- a/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
@@ -146,7 +146,8 @@ namespace Xamarin.PropertyEditing.Mac
 				// Also update the expander state. This is needed when the property editor first
 				// displays and category groups default to expanded, to show the right icon
 				NSView view = outline.GetView (0, row, makeIfNecessary: false);
-				if (view?.Subviews != null && view.Subviews.Length > 0 && view.Subviews[0] is NSButton expander)
+				NSView[] subviews = view?.Subviews;
+				if (subviews != null && subviews.Length > 0 && subviews[0] is NSButton expander)
 					expander.State = NSCellStateValue.On;
 			} else if (facade.Target is ObjectPropertyViewModel ovm) {
 				NSView view = outline.GetView (0, row, makeIfNecessary: false);
--- a/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
@@ -140,9 +140,15 @@ namespace Xamarin.PropertyEditing.Mac
 				return;
 			}
 
-			if (facade.Target is PanelGroupViewModel group)
+			if (facade.Target is PanelGroupViewModel group) {
 				this.dataSource.DataContext.SetIsExpanded (group.Category, isExpanded: true);
-			else if (facade.Target is ObjectPropertyViewModel ovm) {
+
+				// Also update the expander state. This is needed when the property editor first
+				// displays and category groups default to expanded, to show the right icon
+				NSView view = outline.GetView (0, row, makeIfNecessary: false);
+				if (view != null && view.Subviews[0] is NSButton expander)
+					expander.State = NSCellStateValue.On;
+			} else if (facade.Target is ObjectPropertyViewModel ovm) {
 				NSView view = outline.GetView (0, row, makeIfNecessary: false);
 				SetRowValueBackground (view, valueBackground: true);
 			}
--- a/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
+++ b/Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
@@ -146,7 +146,7 @@ namespace Xamarin.PropertyEditing.Mac
 				// Also update the expander state. This is needed when the property editor first
 				// displays and category groups default to expanded, to show the right icon
 				NSView view = outline.GetView (0, row, makeIfNecessary: false);
-				if (view != null && view.Subviews[0] is NSButton expander)
+				if (view?.Subviews != null && view.Subviews.Length > 0 && view.Subviews[0] is NSButton expander)
 					expander.State = NSCellStateValue.On;
 			} else if (facade.Target is ObjectPropertyViewModel ovm) {
 				NSView view = outline.GetView (0, row, makeIfNecessary: false);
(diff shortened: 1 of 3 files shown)


Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -140,9 +140,15 @@ public override void ItemDidExpand (NSNotification notification)
...
 			}
 
-			if (facade.Target is PanelGroupViewModel group)
+			if (facade.Target is PanelGroupViewModel group) {
 				this.dataSource.DataContext.SetIsExpanded (group.Category, isExpanded: true);
-			else if (facade.Target is ObjectPropertyViewModel ovm) {
+
+				// Also update the expander state. This is needed when the property editor first
+				// displays and category groups default to expanded, to show the right icon
+				NSView view = outline.GetView (0, row, makeIfNecessary: false);
```
Therzok, line 148: I don't think this is the right way to do it, you need to use `expandItem:` https://developer.apple.com/documentation/appkit/nsoutlineview/1532826-expanditem?language=objc

BretJohnson: The intention here isn't to expand the item. It's already been expanded, which is why ItemDidExpand is being called. This code is only there to update the icon so that the icon shows the right state (the expanded icon instead of the collapsed icon). Without it the icon is wrong when the item is expanded initially.

netonjm: @BretJohnson this is because you are using a custom expander? just to be in context

BretJohnson: @netonjm I'm not sure if you'd call this a custom expander - I didn't write this code originally. Would you mind taking a look at the code to see. The expand/collapse NSButton in the NSOutlineView looks to just be the default one, not particularly custom. However, there is lots of code there already to update the button State, so that must be needed; the code was just missing in the default case, where the item defaults to being expanded instead of the user explicitly expanding.

netonjm: A question I have is if ItemDidExpand needs this change ItemDidCollapse will need another patch in the same way?

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -140,9 +140,15 @@ public override void ItemDidExpand (NSNotification notification)
...
+				if (view != null && view.Subviews[0] is NSButton expander)
```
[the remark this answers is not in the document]
BretJohnson: I'm not certain & the code above doesn't handle this case. But that said, I added a check just to be safe.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
BretJohnson: No. The icon defaults to collapsed, so when that happens by default the case is covered and if the user explicitly collapses, that's also covered by the existing code.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
```diff
@@ -140,9 +140,15 @@ public override void ItemDidExpand (NSNotification notification)
...
 
-			if (facade.Target is PanelGroupViewModel group)
+			if (facade.Target is PanelGroupViewModel group) {
 				this.dataSource.DataContext.SetIsExpanded (group.Category, isExpanded: true);
-			else if (facade.Target is ObjectPropertyViewModel ovm) {
+
+				// Also update the expander state. This is needed when the property editor first
+				// displays and category groups default to expanded, to show the right icon
+				NSView view = outline.GetView (0, row, makeIfNecessary: false);
+				if (view?.Subviews != null && view.Subviews.Length > 0 && view.Subviews[0] is NSButton expander)
```
Therzok, line 149: Please don't query Subviews 3 times. It allocates an array on every pinvoke.
```suggestion
				if (view.Subviews is NSView[] subviews && subviews.Length > 0 && subviews is NSButton expander)
```

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
Therzok: I'm still not sure how we can get into a state where the outline view state and the UI are not synced.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
BretJohnson: Updated.

Xamarin.PropertyEditing.Mac/PropertyTableDelegate.cs
BretJohnson: In this code, the icon for expand/collapse state is always set explicitly. The icon UI would never change if the code didn't update it. It must be a custom thing, where automatically updating the UI just doesn't happen otherwise. I'm not sure why that is as I didn't write this originally, but I expect it was done on purpose out of necessity. Given all that, the fix here is to explicitly set the icon in the one case that's not handled today - when it's expanded by default. After that, if the user expands/collapses the event notifications update the icon (via existing code in that class).

netonjm: https://github.com/xamarin/Xamarin.PropertyEditing/pull/520/commits/96561f19bf1c81e24c852e0a88ef46bd72e05e38

Therzok: That's a lot of code for things that have equivalent APIs. It may not have been the case in the past, but this can be done properly with less perf overhead and overall code.

BretJohnson:
@netonjm @Therzok - OK, I followed Jose's suggestion, removing the custom Catalina code and adding a call to expand groups that should be expanded by default (auto expanded). My testing on Monterey looks good. Can you please review/approve. Thx.

Therzok (approved):

PR #830, opened 2023-03-28
Fix API not supported on Catalina

BretJohnson:
NSTableView.FullWidth is only explicitly supported on Big Sur and later, so don't set it for older versions of macOS. For older versions of macOS, FullWidth is apparently the default so doesn't need to be specified.

This updates the layout UI tweaks included in
https://github.com/xamarin/Xamarin.PropertyEditing/pull/817.

Fixes AB#1776438

### DIFF
--- a/Xamarin.PropertyEditing.Mac/MacSystemInformation.cs
+++ b/Xamarin.PropertyEditing.Mac/MacSystemInformation.cs
@@ -1,0 +1,48 @@
+using System;
+
+namespace Xamarin.PropertyEditing.Mac
+{
+	public class MacSystemInformation
+	{
+		public static readonly Version Ventura = new Version (13, 0);
+		public static readonly Version Monterey = new Version (12, 0);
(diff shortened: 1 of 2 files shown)


mrward (approved):
This looks good to me. I did notice that in some places in VS Mac we use a simpler way to determine the macOS version, at least if the code is running on .NET.

https://github.com/xamarin/vsmac/blob/a41f0d0659ec0e3d097bfbbcd0246da10acf76d7/main/src/core/Impl/MainToolbarImpl/AwesomeBar.cs#L74

https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.ismacosversionatleast?view=net-7.0

BretJohnson:
@mrward Good to know on the `OperatingSystem.IsMacOSVersionAtLeast` OS version check. Thanks. Here I think I'll stick with using `MacSystemInformation` because it seems to be the most common (for better or worse) method used currently.

BretJohnson:
@gitbot backport d17-6

PR #832, opened 2023-04-20
Update to use new, shared compliance template

BretJohnson:


japarson (approved):
I suggest setting up a .gdnsuppress file for potential use in the future.

BretJohnson:
It turns out there's no immediate driver to use the shared ompliance template, given that I ended up not using it for policheck, binskim, or antimalware (all of which had issues when used from this repo). Closing.

PR #845, opened 2024-09-05
Add notification in README

BretJohnson:
Add the text "“This repository is no longer actively maintained or developed.” to the README for this public repo. UITools is the last remaining client that uses this code & it now embeds the source in its repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,4 @@
 ### Build Status
 [![Build Status](https://dev.azure.com/xamarin/public/_apis/build/status/xamarin/Xamarin.PropertyEditing/Xamarin.PropertyEditing%20(Public)?branchName=main)](https://dev.azure.com/xamarin/public/_build/latest?definitionId=51&branchName=main)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,4 @@
 ### Build Status
 [![Build Status](https://dev.azure.com/xamarin/public/_apis/build/status/xamarin/Xamarin.PropertyEditing/Xamarin.PropertyEditing%20(Public)?branchName=main)](https://dev.azure.com/xamarin/public/_build/latest?definitionId=51&branchName=main)
```
spadapet, line 2: Would it be reasonable to remove this "Build Status" indicator? I assume the build pipeline would be disabled or removed.

spadapet (approved):

README.md
BretJohnson: We're still going to use this for older UITools branches, so we'll need to keep the pipelines and I think there's a little bit of value in keeping the build status indicator here.

